PR #5, opened 2021-11-09
add intro to r/rstudio

pm0kjp:


pm0kjp:
Canceling this in favor of better options

PR #10, opened 2021-11-12
QA instructions

pm0kjp:
Aside -- I wasn't sure where to stick images, so I created a media folder at the root level to use for root level .md documents, but I could be convinced otherwise...


leemc-data-ed:
I wonder if it might be worth adding a line suggesting that folks take a look at previous QA issues if they're confused about anything (since we will have completed the first few ourselves and presumably done it correctly).

quality_assurance.md
```diff
@@ -0,0 +1,93 @@
...
+
+When a module creator is ready to request that their module be included, they will create a Pull Request (PR).  This begins the work of quality assurance.  As someone who is reviewing the modules created by others, it's important to have use a consistent method for evaluating content.
+
+## Step 1: Create an Issue
+
+* Click on "Issues" or go to https://github.com/arcus/education_modules/issues.
+* Choose "New Issue" (or if your screen is small, just "New").  This is a green button on the right side:
+![Screenshot of where to click for a new issue](media/new_issue.png)
+* Give your issue a good title: "QA" plus the proposed directory name from the PR. For example, if the PR includes a new module with the directory named "reproducibility", the title would be "QA reproducibility".)
+* In the "Write" tab area, paste the following (from `# Module Quality Assurance Report for PR #[put in the PR number here] ` to `* [ ] description or quote, line ___ in file ____`).  In square brackets, remove the square brackets and their contents and replace with the appropriate values.
```
leemc-data-ed, line 11: I think it might be clearer if it said "Where there are square brackets," rather than "In square brackets".

leemc-data-ed, line 11: I also think we could mentioned the pinned sample issue here.

leemc-data-ed (changes requested):
Just a couple of small changes, otherwise lgtm!

pm0kjp:
> @pm0kjp I wonder if it might be worth adding a line suggesting that folks take a look at previous QA issues if they're confused about anything (since we will have completed the first few ourselves and presumably done it correctly). Or point them to the pinned sample I just saw.

Good idea!  I mentioned both the sample issue and a "complete" (I mean it isn't yet but it will be) QA review in my latest commit

quality_assurance.md
pm0kjp: Great suggestion!  Included it and added a link to a real issue in a recent commit.

pm0kjp: done

pm0kjp:
OK I added a couple more commits to bring things a bit more up to snuff, plus I added a new section called "Structural Elements" which I realized I wanted/needed when I started actually doing a review!

pm0kjp:
Also added new images, including logo and favico

pm0kjp:
OK, re-asking for approval -- @leemc when you get a chance (does NOT have to be over the weekend, I'm just making up for lost time) do click "approve" (don't just give a comment)

leemc-data-ed:
Just found one spot where some text got duplicated, will fix and approve!

pm0kjp:
OK, the whole team should have admin privileges now!

leemc-data-ed (approved):
Lgtm!

PR #28, opened 2021-12-06
Template update

rosemm:
Updating the sample template as per our decisions last week (e.g. dropping the Contents section from overview), and also showing how to use YAML for short description, long description (is this module right for me? bit) and learning objectives. This means learning objectives don't need to be copy-pasted at the bottom of the module anymore, in the feedback section. Yay!


leemc-data-ed (approved):
Thanks for doing this! Lgtm

PR #32, opened 2021-12-13
update gitignore

rosemm:
updating gitignore so it will ignore ipynb checkpoints in subdirectories as well


leemc-data-ed (approved):
Awesome, thanks!

PR #51, opened 2022-03-08
Joy update reproducibility

pm0kjp:
This was already approved, but I've updated the module to meet newer standards.  Can you just give this a quick one-over, @drelliche?   Not the full QA script, just flipping through to see if there's anything glaring. I'm seeing some old styles come through but I think it's just the Dreaded Caching....

Aside, we should think about what a "periodic update / improvement" QA looks like for already-approved modules.


drelliche (approved):
Looks good rendered!

PR #54, opened 2022-03-09
Template and documentation updates

franzenr:
Updates to templates and other various documentation

Highlights include: 
* creation of versioning guidelines
* adding version numbers to module template as well as QA checklist template
* splitting QA md into separate guide and template files
* improvements to QA such as better alt text guidance, clearer operationalization of items, reducing double-barreled items, new organizational structure, etc 
* removing QA template text from how_to.md (to avoid multiple versions floating around) and some minor changes

More specific details available in each commit message


pm0kjp (approved):
These are all fantastic changes that really aid in module creation!  Thank you!

drelliche (approved):
Looks great to me!

As a one-time thing, should modules currently in QA close those issues and start a new issue with the new (way clearer!) QA template?

quality_assurance_template.md
```diff
@@ -0,0 +1,107 @@
...
+
+
+## Code availability
+If the module includes code that learners may want to run:
+* [ ] A link is provided to run the code in binder
+* [ ] A link to the raw code and any other necessary materials (e.g., data) is provided for learners who prefer to work on their own machine
+
+## Formative assessment
+
+* [ ] Frequent [formative assessment](https://carpentries.github.io/instructor-training/02-practice-learning/#identifying-and-correcting-misconceptions) in the form of knowledge checks (referred to as quizzes) and/or hands-on exercises
```
rosemm, line 67: to avoid confusion, maybe just "in the form of quizzes and/or hands-on exercises"?

rosemm (approved):
Look great! Thanks for putting this together!

quality_assurance_template.md
franzenr: yeah good point! I think I was trying too hard to not totally change previous language, but that doesn't seem to be a sensible thing to do in this instance!

franzenr:
Okay, I think we're finally good to merge this into main now! 

In the end, since we have a lot of writing going on currently, the decision for modules mid-QA is that it will be left up to the reviewer whether they want to
(1) transfer to the new QA template before proceeding
OR
(2) finish under the version of the QA process they started with, adding that module to the list of already merged modules that will need to be updated later (at some point after pilot launch).

PR #64, opened 2022-03-16
Update hint and consolidate .css

pm0kjp:


pm0kjp:
@rosemm most of these changes are already actually live because I somehow updated the .css on AWS and didn't push my changes through github!  Eeek!  So although this looks like a radical rewriting, the only real change is to the hint box.  If you can just approve on my word of honor that would be amazeballs

rosemm (approved):
lgtm

PR #82, opened 2022-03-24
Check statistical tests in open source software module

rosemm:
Bring up to date with new module template; fix some broken links

### DIFF
--- a/statistical_tests/statistical_tests.md
+++ b/statistical_tests/statistical_tests.md
@@ -2,7 +2,7 @@
 
 author:   Rose Hartman
 email:    hartmanr1@chop.edu
-version:  1.1.0
+version:  1.0.1
 module_template_version: 2.0.0
 language: en
 narrator: UK English Female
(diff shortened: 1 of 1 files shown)


statistical_tests/statistical_tests.md
```diff
@@ -2,13 +2,15 @@
 
 author:   Rose Hartman
 email:    hartmanr1@chop.edu
-version:  1.0.0
+version:  1.1.0
```
leemc-data-ed, line 5: I wonder if this would even be considered version 1.0.1? Seems like most of the changes are behind the scenes, though that one additional external resource link might bump it up...

leemc-data-ed (approved):
Just the question about the version... But I don't feel strongly enough about it not to approve ;)

statistical_tests/statistical_tests.md
rosemm: You're probably right! I changed that at the beginning because I was anticipating much more substantial updating, like I had to do for ggplot2, but this is almost all on the order of fixing broken links. I'll revise to 1.0.1

PR #85, opened 2022-03-24
Joy bug squash r basics intro (Fix #83)

### LINKED ISSUE #83: R Basics: Introduction
I was copying over content from this module to update the ggplot2 module and I noticed a few little errors: 

- [x] line 186: missing alt text
- [x] line 240: reference to wrong folder for code
- [x] line 255: reference to wrong folder for code
- [x] line 262: reference is to branch (and its the visualize branch as well, not the intro one), so I think it will break if/when that branch is deleted, although it's still working now
- [x] the png and gif referenced in line 262 appear to not be in the media folder on main
- [x] line 276: I think the last file name should be `transform_exercises.rmd` rather than `transform.rmd`
- [x] line 671: reference to wrong folder for code

pm0kjp:
Fix #83


pm0kjp:
@rosemm @franzenr I'm pretty sure this will do the trick :)

rosemm (approved):
LGTM!

PR #90, opened 2022-04-01
Update setup modules

drelliche:
They now have parallel titles, and the language about why we use `main` instead of `master` has been updated.


rosemm (approved):
LGTM! Thanks!

PR #94, opened 2022-04-07
Update directories_and_file_paths.md

leemc-data-ed:
Based on learner feedback, fixed "Writing a file path" quiz so that an incorrect answer is not accepted as correct. Shouldn't take long to check!

### DIFF
--- a/directories_and_file_paths/directories_and_file_paths.md
+++ b/directories_and_file_paths/directories_and_file_paths.md
@@ -54,7 +54,7 @@ script: https://kit.fontawesome.com/83b2343bd4.js
 
 If you've ever saved a file on your computer, you probably have some familiarity with how files and folders work. Maybe you save a file to your Desktop, or in your Documents folder, or perhaps you made a special folder to hold all of your files relating to a specific project. In some ways, it works very similarly to a physical filing system; related files go in labeled folders, which are stored in a known location (a file cabinet in a file room somewhere).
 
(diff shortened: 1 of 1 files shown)


franzenr (commented):
checked it out -- looks like it works as intended now! 

As an aside, I went through the entire module quickly just because I was curious -- there's so many of our modules I haven't had a chance to read yet, and this gave me the excuse! :P 

In doing so, I noticed that the first quiz question, assesses a synonym for directory, despite never being given (ie, at no point does the educational bit of the content thus far use the word directory). While it's a simple + familiar enough concept that I'm sure most folks can intuit that without it having been mentioned, I wanted to flag it in case that's something you'd want to change :)

Also, the only place in the module that drive or volume are defined is in the answer explanation text. Those are tangential terms to the primary focus of the module, so maybe that's not a problem, but if we think it's useful for learners to have that clear disambiguation, I wonder if we might want to add that information to the body of the material somewhere? I know as a learner, I'm guilty of not reading the explanations in full if I answered the question right on the first try (perhaps that's a personal fault though haha)

leemc-data-ed:
@franzenr Good call! I explicitly call out "directory" as a synonym for "folder". My inclination is to leave volume and drive just in the answer to that question, since they aren't super important for the module, and I want to keep the scope narrow. If no one really reads it I think that's fine.

rosemm (approved):

PR #117, opened 2022-05-27
R reshape long wide

pm0kjp:
Hi @rosemm I figured we could trade QAs?  I am pointing the animated gif on the "get your R environment set up" page to main, even though this isn't on main yet.  So it looks broken, but it's not, I promise :).


rosemm:
Sounds good! I'll hop on it now :)

r_reshape_long_wide/r_reshape_long_wide.md
```diff
@@ -0,0 +1,668 @@
...
+* a single column that holds values.
+
+Let's start by thinking about what constitutes a unique observation.  
+
+* Does a `subject_id` alone constitute a unique observation?  No, because two subjects have more than one biosample collection.  
+* What about the `subject_id` along with date?  Again, no, because we have one subject with two biosample collections on the same day.
+* `subject_id`, date, and time?  That would work... but we now have three columns required to make a unique observation.  Is there a better way?
+* What about the `biosample_id`?  Yes, that would work.  Each sample (blood or saliva) has its own `biosample_id` and those ids don't repeat.
+
+OK, so we'd like our  "one or more columns that uniquely identify and observation" to be the `biosample_id`.  One problem is that we have two sets of these -- one with the suffix `_1` and one with the suffix `_2`.  We'll have to do some clean up, and we can do that when we cut our data frame down the middle.
```
rosemm, line 599:
```suggestion
OK, so we'd like our  "one or more columns that uniquely identify an observation" to be the `biosample_id`.  One problem is that we have two sets of these -- one with the suffix `_1` and one with the suffix `_2`.  We'll have to do some clean up, and we can do that when we cut our data frame down the middle.
```

r_reshape_long_wide/r_reshape_long_wide.md
```diff
@@ -0,0 +1,668 @@
...
+
+At this point, you might be thinking, "Hey!  If we just stack `biosample_first` and `biosample_second`, we'll be done, and have some tidy data!".  You're not wrong, and you might be tempted to do just that and not bother with transforming wide data to long data.  Just remember that once you get data into a long format, it's very versatile, and it's worth the effort to put your data into a long format first.
+
+Go to lines 230 - 265 and read through the explanations and run the code in this section of the file.
+
+By the end of this section, you should have a `biosample_data_long` object that has our data in long format.
+
+### Pivoting Long to Wide
+
+By now, you should have long data that's ready to be pivoted to wide -- but this time, it will be wide data that's tidy, a vast improvement over the data we began with.
```
rosemm, line 644: I recommend moving the "we would like data that looks like this" table from the rmd file to this file, because reading the raw md table format will be hard for folks that aren't used to it. Here it will render as a nice table.

pm0kjp: Good call good call!

r_reshape_long_wide/r_reshape_long_wide.md
```diff
@@ -0,0 +1,668 @@
...
+```
+# Simplest case where column names are character data
+relig_income
+relig_income %>%
+  pivot_longer(!religion, names_to = "income", values_to = "count")
+```
+
+Let's try that in R!  In your R environment, open the file `reshape_data_exercises.Rmd` from within the `r_reshape_wide_long` folder.
+
+Please read over the first 70 lines of code and execute that code in the R Markdown document.  This is what you'll be doing:
```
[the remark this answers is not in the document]
pm0kjp: Agree!  I can put a screenshot in, too.

r_reshape_long_wide/r_reshape_long_wide.md
```diff
@@ -0,0 +1,668 @@
...
+`values_drop_na` will be used: `FALSE`, don't drop empty (NA) values.  But if you want to change these (or any other) optional arguments, you can.
+
+So, what do we put in as our `data`, and what do we add for `cols`?  Look a bit further down in the help file under the **Arguments** section and you'll see that `data` refers to the data frame you want to reshape (pivot), and `cols` refers to the columns you want to pivot.
+
+When it comes to choosing columns, you generally want to pivot all the columns of your data frame except for the ones that uniquely identify an observation (a group, a patient, a vital signs reading, an environmental sensor reading at a particular moment, a town).  Sometimes that's just one column, as we'll see in our first example.  Other times, you might need two or more columns to uniquely identify an observation (like patient MRN and encounter ID to uniquely identify a patient encounter).
+
+This might all seem very theoretical and hard to understand.  That's why normally we like to scroll down in the help file all the way to the bottom, where there are some examples.  Let's start with a simple case.  This is the example given in the help file:
```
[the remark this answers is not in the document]
pm0kjp: not a bad idea, I'll add!

rosemm (approved):
Woo hoo let's do it!

PR #133, opened 2022-06-01
Revisions r basics

franzenr:
Hi Joy, I'm putting you on this one since these are your modules! 
In particular, I had to do a little rearranging section titles/structure for the transform module -- make sure you like how it looks, feel free to add an introductory section to the hands on if desired like you did in viz. 

The environment setups should now all be the same (including alt text), setup only appears once (at beginning) in transform, and I also noticed a few decorative images not handled appropriately in terms of alt text, so I did those really quickly as well.


pm0kjp (approved):
TYSM!  Great improvements all around

Issue #166, opened 2022-07-11
Quiz marking correct answer wrong


drelliche:
In this question:

> You create a new folder titled canines in learning_bash-main. What command would you use to move the files dog.txt and wolf.txt into the canines folder? Use the relative locations of the files and folders.

the answer `mv dog.txt wolf.txt canines` is marked wrong.

drelliche:
Closed by merging PR https://github.com/arcus/education_modules/pull/167

PR #167, opened 2022-07-11
trim whitespace from quiz input

drelliche:
In response to Issue https://github.com/arcus/education_modules/issues/166 , the quiz questions were not trimming extra whitespace before checking if the answer was correct. I believe this will fix that.

### DIFF
--- a/bash_command_line_102/bash_command_line_102.md
+++ b/bash_command_line_102/bash_command_line_102.md
@@ -225,11 +225,11 @@ The `find` function searches the names of all folders and files in the current d
 ***
 
 
-2. What you command will return all files with a `.dat` ending that contain the word `Weight`
+2. What you command will return all files with a `.dat` ending that contain the word `Weight`?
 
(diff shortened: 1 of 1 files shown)


bash_command_line_102/bash_command_line_102.md
```diff
@@ -225,11 +225,11 @@ The `find` function searches the names of all folders and files in the current d
 ***
 
 
-2. What you command will return all files with a `.dat` ending that contain the word `Weight`
+2. What you command will return all files with a `.dat` ending that contain the word `Weight`?
```
rosemm, line 228:
```suggestion
2. What command will return all files with a `.dat` ending that contain the word `Weight`?
```

rosemm: @drelliche extra word here, right?

rosemm (approved):
This appears to be working correctly now!

PR #191, opened 2022-08-12
Data storage models fix

### LINKED ISSUE #165: Content for quiz question appears after the question itself
Quiz: Storage on your Computer - What is true of desktop storage? Option 3 relies on knowing content presented later (under "on premises network"). Need to rework or reorder so learners are equipped to answer the question at the point where it's presented.

### LINKED ISSUE #173: Quiz Question Mismatch
For the **_Quiz: Storage on your Computer_**, the answers do not relate to the question nor the content being tested. The answers are from the quiz, _**Quiz: On-Premises Network**_, and the drop-down answer box is the same as well. I think there may be a mistake there.

### LINKED ISSUE #190: "Cost of storage in the cloud is inversely related to how instantly it can be accessed". User suggests we replace "instantly" with "how quickly", "how long it takes to access data" or "the speed with which data may be accessed"

rosemm:
Tweaking wording in the cloud storage quiz (fixes #190), and combining local and network quizzes together (fixes #165, fixes #173). 
It was originally designed as two "parallel" quiz questions, with the same set of answers but first applying to local storage and later to network storage (and they had the same solution because the first was "which of these is true" and the second was "which of these is NOT true"). That wasn't working well because the repeated content was potentially confusing (user report #173 ), and because several of the answer options were not actually explained before the local storage quiz, only before the later network storage quiz (#165 ). I reorganized the content to make it one matrix quiz after the network storage section, rather than two separate multiple choice questions.

### DIFF
--- a/data_storage_models/data_storage_models.md
+++ b/data_storage_models/data_storage_models.md
@@ -2,7 +2,7 @@
 
 author:   Nicole Feldman
 email:    feldmanna@chop.edu
-version:  1.0.0
+version:  1.1.0
 language: en
 narrator: UK English Female
 title: Types of Data Storage Solutions
--- a/data_storage_models/data_storage_models.md
+++ b/data_storage_models/data_storage_models.md
@@ -174,14 +174,14 @@ For all its computational advantages, on premises network storage has significan
 
 ### Quiz: Cloud Storage
 
-True or False: Cost of storage in the cloud is inversely related to how quickly data can be accessed.
+True or False: Cost of storage in the cloud is directly related to how quickly data can be accessed.
 
-[( )] TRUE
-[(X)] FALSE
+[(X)] TRUE
+[( )] FALSE
 ***
 <div class = "answer">
 
-The most economical storage in the cloud is coldline or archival tier storage. This option can offer up to a several day retrieval time.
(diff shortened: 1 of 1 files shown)


rosemm:
@nafeldman Hi Nicole! I added you to this pull request since you're the original author on the module (this is the Data Storage Models one), and I wanted to keep you in the loop. We're happy to handle the review etc. though, so you don't have to check it over unless you want to.

data_storage_models/data_storage_models.md
```diff
@@ -143,20 +128,25 @@ Both optical discs and USB flash drives require playback media and can suffer ir
...
+- [ [X]      [ ] ] Once you have made the resource investment, making changes is quite easy
+- [ [ ]      [X] ] Files can be only accessed over the network or through a VPN client
+-
 
-[( )] Requires physical space to maintain and host servers
-[( )] Requires in house IT staff or vendor support to maintain the technical infrastructure
-[(X)] Once you have made the resource investment, making changes is quite easy
-[( )] Files can be only accessed over the network or through a VPN client
 ***
+
```
drelliche, line 143: answer box is appearing before answer is submitted

drelliche (approved):

Issue #193, opened 2022-08-16
Rename folders and files


franzenr:
All git modules so far (including the [windows setup module](https://github.com/arcus/education_modules/tree/main/git_setup_windows) that complements this mac/linux module) are named "git_*something something something*"

May want to consider updating this module to follow the same naming structure for easier reference/location/linking -- at minimum, perhaps we want both of the setup modules named the same (meaning we could choose instead to adapt the windows module to have similar file and directory naming as this module instead). 

Note that if we update, other modules that link out to this module will need to be updated with the new link.

franzenr:
@drelliche assigning this to you since you wrote the Windows module, and Rose H is of course out on leave

drelliche:
@franzenr This turned out to be pretty quick and easy with a find and replace! Take a look at PR #194

PR #194, opened 2022-08-16
rename module to git_setup_mac_and_linux

drelliche:
#193


franzenr (changes requested):
One tiny pedantic change request since we're already here -- YAML title calls this Setting Up Git _for_ Mac and Linux, while displayed title is *on* Mac and Linux -- can you fix to be on so it agrees with YAML + windows module? 

I'd update it myself with a code suggestion, but it's just outside of the range of impacted lines where it will let me insert a code suggestion via github.com

drelliche:
Good catch @franzenr! All 4 places (yaml and title for both Mac/linux and Windows) now use "on"

franzenr (approved):

franzenr:
🙌

Issue #213, opened 2022-11-02
man command does not work on git bash


franzenr:
record 122: man command does not work on git bash, returns "command not found" error

franzenr:
you know the drill -- assigning to you since you're co-author, feel free to redistribute as necessary

PR #216, opened 2022-11-02
add warning about no man in GitBash

drelliche:
This addresses Issue https://github.com/arcus/education_modules/issues/213

The `man` command is not enabled in GitBash (and from the Git Bash conversation each time the issue is raised, is unlikely to every be available). This warning tells GitBash users that they can get the same info from an internet search.

I considered introducing `--help` here, but rejected it as it is far less helpful and has far less information than the manual, and the information provided is likely to be quite confusing for new users of Bash.

### DIFF
--- a/bash_command_line_101/bash_command_line_101.md
+++ b/bash_command_line_101/bash_command_line_101.md
@@ -488,9 +488,15 @@ The quotes around the text can also be safely omitted in this case, but that isn
 
 - Software Carpentries offers an open source course [introducing new users to the Unix Shell](https://swcarpentry.github.io/shell-novice/)
 
-### The bash manual
+### Getting help from the manual
 
-The bash manual is available to you whenever you are in a bash shell.  Entering `man command_name` reveals the documentation (`man` is short for "manual") for the function `command_name`.  This can be useful if you think a command might be useful and want to learn more about it, or have forgotten how to use a command or what its options are.
+<div class="warning">
+
+The bash manual is available to you in most bash shells, although not in Git Bash. If you are using Git Bash, an internet search for "`command_name` bash manual" will usually get you the same information.
+
+</div>
+
+Entering `man command_name` reveals the documentation (`man` is short for "manual") for the function `command_name`.  This can be useful if you think a command might be useful and want to learn more about it, or have forgotten how to use a command or what its options are.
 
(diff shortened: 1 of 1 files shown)


drelliche:
@franzenr you seem to be on a GitHub roll today, and this should be quite short, but can certainly go to someone else.

franzenr:
This looks good to me, although this reminds me that one thing I noticed when looking through these modules myself is that the way we're presenting this (that is, teaching them how to use a specific command to get additional help, rather than just linking them to some outside resource), it should probably be included in the actual module content itself, not nestled under Additional Resources. 

I had been planning to drop this in a new issue at some point today -- I can quickly draw that one up if you'd like, and we can get them both taken care of with one PR? @drelliche

drelliche:
@franzenr I made the manual it's own level 2 section before additional resources.

bash_command_line_101/bash_command_line_101.md
```diff
@@ -480,17 +480,15 @@ The quotes around the text can also be safely omitted in this case, but that isn
...
-- [Reinforce Your New Knowledge through this Learning the Shell Page](https://linuxcommand.org/lc3_learning_the_shell.php)
+The bash manual is available to you in most bash shells, although not in Git Bash. If you are using Git Bash, an internet search for "`command_name` bash manual" will usually get you the same information.
 
-- Software Carpentries offers an open source course [introducing new users to the Unix Shell](https://swcarpentry.github.io/shell-novice/)
-
-### The bash manual
+</div>
 
-The bash manual is available to you whenever you are in a bash shell.  Entering `man command_name` reveals the documentation (`man` is short for "manual") for the function `command_name`.  This can be useful if you think a command might be useful and want to learn more about it, or have forgotten how to use a command or what its options are.
+Entering `man command_name` reveals the documentation (`man` is short for "manual") for the function `command_name`.  This can be useful if you think a command might be useful and want to learn more about it, or have forgotten how to use a command or what its options are.
```
franzenr, line 491:
```suggestion
Entering `man command_name` reveals the documentation for the function `command_name`  (`man` is short for "manual").  This can be helpful if you think a command might be useful and want to learn more about it, or have forgotten how to use a command or what its options are.
```
1. Improve sentence flow
2. Avoids using useful 2x in one sentence

franzenr:
one quick rewrite suggestion and then I think we're good to go! 🙌  @drelliche

franzenr (approved):

PR #241, opened 2022-12-02
Highlight box update

rosemm:
This is an update to the css file for all modules. It make several adjustments to highlight boxes:

- Adds new highlight boxes (e.g. `cool-fact`, `history`) 
- Replaces the icons before several kinds of highlight boxes
- Adds automatic text to the beginning of each highlight box
- Renames `learnmore` to `learn-more` for consistency with other multi-word divs
- Removes `hint` boxes, which we were not using
- Update from Font Awesome 5 to 6

In addition to the css file itself, this PR also includes changes to the module template. The section on highlight boxes was substantially updated both to add the new kinds of boxes available, as well as to clarify the purpose of existing highlight boxes:
- The instructions for when to use each kind of highlight box have been changed
- The examples for each highlight box have also been updated

NOTE: This qualifies as a major revision of the module template. Modules made with the previous template may use highlight boxes in a way that would be inconsistent with the updated template, especially ones that were particularly unclear or inconsistent in the previous template (e.g. care vs. help, warning vs. important, options vs. learnmore). 

Finally, all existing markdown files that included the string `learnmore` have had it updated to `learn-more`. Oh, and I alphabetized the list of highlight boxes, both in the module template and the css file, to make it a little easier to navigate.

### DIFF
--- a/a_sample_module_template/a_sample_module_template.md
+++ b/a_sample_module_template/a_sample_module_template.md
@@ -115,14 +115,14 @@ It has a gears icon and always begins with the text "Behind the scenes".
 For example:
 
 <div class = "behind-the-scenes">
-**Behind the scenes**
+<b style="color: rgb(var(--color-highlight));">Behind the scenes</b><br>
 
 The commit number is a hash and associated details
 
 </div>
 
 <div class = "behind-the-scenes">
-**Behind the scenes**
+<b style="color: rgb(var(--color-highlight));">Behind the scenes</b><br>
 
 In R the `<-` and `=` can both be used for assignment because...
 
@@ -135,14 +135,14 @@ It has a hand-holding-heart icon and always begins with the text "A little encou
 For example:
 
 <div class = "care">
-**A little encouragement...**
+<b style="color: rgb(var(--color-highlight));">A little encouragement...</b><br>
 
 This is a topic with a tremendous amount of jargon, which can make resources you may find online hard to understand for folks new to the field. When that happens it's easy to feel like there's something wrong with you if you don't get it, but that's not the case! Those kinds of gatekeeping explanations are a failure on the part of the writer, not the learner.
 
 </div>
 
 <div class = "care">
-**A little encouragement...**
+<b style="color: rgb(var(--color-highlight));">A little encouragement...</b><br>
 
 Feeling overwhelmed? It takes a long time to learn git, so don't be disheartened if it doesn't click initially. Just focus on stage, commit, and push. Ignore the rest for now, until you've had a chance to practice just the stage-commit-push process several times.
 
@@ -155,7 +155,7 @@ It has a brain icon and always begins with the text "Did you know?"
 For example:
 
 <div class = "cool-fact">
-**Did you know?**
+<b style="color: rgb(var(--color-highlight));">Did you know?</b><br>
 
 Functions like this are sometimes called "syntactic sugar" because they don't change anything about how the code runs, they just make it easier for humans to read, the way that sugar makes food sweeter without adding any nutrition.
 
@@ -163,7 +163,7 @@ Functions like this are sometimes called "syntactic sugar" because they don't ch
 
 
 <div class = "cool-fact">
-**Did you know?**
+<b style="color: rgb(var(--color-highlight));">Did you know?</b><br>
 
 This is a box showing how images work in a highlight box.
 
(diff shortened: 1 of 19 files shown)


drelliche:
@rosemm I'm seeing an issue with the `before` text showing up before each item in a list inside of a box. 

Take a look at this box in the [Bash: Combining Commands module](https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/highlight_box_update/bash_103_combining_commands/bash_103_combining_commands.md#7)
Screen Shot 2022-12-02 at 11 19 46 AM

rosemm:
Dagnabbit! Good catch --- I'll see if I can get the lists to behave properly. Thanks!

drelliche:
Later in the [same module](https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/highlight_box_update/bash_103_combining_commands/bash_103_combining_commands.md#7) I'm also seeing the title showing up more than once

Screen Shot 2022-12-02 at 11 22 31 AM

drelliche:
For the double title, perhaps this is an issue of me trying to put a bold title in the box from earlier? I can't see any other possible causes of the double `before`.

drelliche:
I've rejected my previous thought about opening with bold doubling the title, because that doesn't happen [here](https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/highlight_box_update/bash_103_combining_commands/bash_103_combining_commands.md#5): 
Screen Shot 2022-12-02 at 11 29 23 AM

drelliche:
Here is another weird effect in [Bash 101](https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/highlight_box_update/bash_command_line_101/bash_command_line_101.md#12)

Screen Shot 2022-12-02 at 11 42 17 AM


The markdown for this (which I'm realizing needs to be updated with the actual link... oops) is:

<div class = "learn-more">
Even though the words "plain text" in the previous paragraph have some non-text attributes, they are all recorded as plain text in the Markdown (`.md`) file of this lesson. If you open the [Markdown file](link/to/git/of/this/file), you will see only characters that can be typed by a keyboard:

```
[**plain text**](https://en.wikipedia.org/wiki/Plain_text)
```
</div>

drelliche:
Also, I very much doubt that this is an issue connected to `learn-more` in particular. Rather this is coming up here because the bash modules show up first alphabetically and I used `learn-more` as a giant catch-all.

rosemm:
Okie dokie, @drelliche this is ready for you again. I did some testing on using lia macros to insert the text for highlight boxes, and it worked for some but I couldn't get it working properly for boxes that contained lists (why are lists so hard?). So I'm officially giving up on automatic text and have instead put in the text for each box in the template, to be copy-pasted along with the div code. Not ideal, but it works. 
I did a more thorough job of testing this as well --- I looked at two bash modules, data storage, and ggplot, and they all rendered correctly (I would expect they would, since the only real change to the css now is updating the icons).

drelliche:
@rosemm Two questions: 
1. Would it make sense to do a find and replace to put the title text on each box? Or will that screw up the boxes that already have titles of some sort and we should probably each go through our own modules updating the boxes and titles?
2. Is it possible to make the title text the same color as the icon? That would make it visually clear that the text is part of the box itself, and not the contents, like in this example:
205339789-03fa7103-3cb3-4dd0-a324-53ced0c82687

rosemm:
For 1, I had the same thought, with the same concern. I think, just glancing at the highlight boxes in some of our existing modules, we'll need to go over them manually to check that they're the right kind of box since we've tweaked the expectations for which boxes get used for what. So I was thinking it probably makes sense to bundle that with adding in that header text (and also maybe removing redundant headers in boxes that already have something there). We could start by doing an automatic text insertion, but I think we'll have to go back and edit anyway in a lot of cases. So maybe it's better to leave them as unchanged as possible for now so we don't insert headers that feel weird given the content that follows in the box?
For 2, excellent idea! Done!

drelliche (approved):

franzenr:
@drelliche @rosemm I hate to be the bearer of bad news, but I think the module template version number was incorrectly incremented. 

If this constituted a major version change, bringing it up from 2.0.1, then the new version should be 3.0.0, not 3.1.0 
whenever a number is incremented, all numbers to right of it should be reset to zero. (See the first bullet under ["Versioning Basics"](https://github.com/arcus/education_modules/blob/main/versioning_guidelines.md)

drelliche:
@franzenr You are so right... ugh, I'm on it.

drelliche:
@franzenr Could you check if this fixes it: https://github.com/arcus/education_modules/pull/243

PR #244, opened 2022-12-21
Update demystifying_python.md

leemc-data-ed:
Updated highlight box (line 274) from "help" to "options." Updated version number and template version number. This seems like the "options" box is clearly the right choice here, though I'm curious if anyone thinks the wording should be tweaked a little.


drelliche (approved):
lgtm!

PR #286, opened 2023-01-25
Small fixes for directories and file paths module

rosemm:
A teeny little PR. Just two small changes:

- fixed a typo I noticed while refreshing myself on the module content
- reworded a quiz question to take out the unnecessary introduction of the word "cohort", which is a technical term people may not be familiar with, and not at all needed for the question


leemc-data-ed (approved):
Lgtm! Thanks!

PR #298, opened 2023-01-26
add how to get solution

pm0kjp:


leemc-data-ed (changes requested):
All of the new content looks good! There are just a couple of issues left (that are actually problems that I left, not ones you introduced)
1. The "Your Task" division in the "Practice Problem" section isn't rendering correctly, but I can't figure out what's wrong with it (but you managed to fix it in the python one, so many you know the trick)
2. If we're capitalizing Google, we'll want to do that in the care box at the end of the "Stuck?" section
But I can make these changes!

leemc-data-ed:
Except I changed the wrong module lol, but I think it's fine, one moment more please.

leemc-data-ed (approved):
Okay, lgtm!

PR #311, opened 2023-02-06
remove reminder to change REDCap link

franzenr:
note reminding to update the REDCap link with the correct module name accidentally made it into the module


leemc-data-ed (approved):
Good catch, thanks!

drelliche (approved):

PR #318, opened 2023-02-15
Quiz typo "easier" to "harder", Fixes #317

### LINKED ISSUE #317: Incorrect info in quiz explanation (Database Normalization)
In the explantation box for the Normalization quiz (pg. 8) in the Database Normalization module, there is the following text: "This repetition means that it's much easier to correct or change data if we try to put as much data as possible into a single table. However, when a normalized model is used, and data is fragmented into concepts, things get easier. Instead of correcting a product name thousands of times, once for each order that product appears on, we only have to correct it once, in the "product" table."
It seems like the first line should actually say the opposite, that it is much **harder** to correct or change data in a large single table. 
This issue is in reference to feedback record ID 224.

pm0kjp:
Also add paragraph breaks and up-version

### DIFF
--- a/database_normalization/database_normalization.md
+++ b/database_normalization/database_normalization.md
@@ -1,7 +1,7 @@
 <!--
 author:   Joy Payton
 email:    paytonk@chop.edu
-version: 1.0.1
+version: 1.0.2
 module_template_version: 3.0.0
 language: en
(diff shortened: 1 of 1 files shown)


leemc-data-ed (approved):
Lgtm!

Issue #212, opened 2022-11-02
first practice problem is confusing


franzenr:
record id: 118
learner reports that the first practice problem was confusing, as there were not examples of it in the content, although there were examples of the opposite scenario. 

(editorializing here: I think what they're referring to is the second "quiz", which is the first time we have learners type out a line of code that is checked)

franzenr:
assigning to you @drelliche since I know you co-wrote w Nicole, and seeing as Nicole is either actively on parental leave or getting ready for it, I figured I'd leave her inbox un-cluttered and you can loop in if necessary. (Also, feel free to pass this on to someone else on the team instead if needed!)

drelliche:
Some possibilities for addressing this are:

- add a hint to the question
- add an additional question before this about navigation that does not require the user to navigate up, only down
- wait and see if any other users report similar confusion
- some combination of these options

My instinct is towards adding a hint, but not an additional question.

@franzenr any strong opinions or other suggestions?

franzenr:
I think adding a hint could be a good idea (perhaps something like "Remember, you can use .. to go up one level in your directory structure") 

Another additional thing that might be useful is to add some specific suggestions of things to play around with when we tell them to "Give it a try" on the Navigation Commands page before the quiz. Since there's no specific prompts of what to do, we don't know whether learners will have tried yet to go back up directories, and therefore maybe didn't fully absorb the information on how to do so. 

Maybe something along the lines of "Give it a try! Use cd, ls, and pwd to explore the directories on your computer. Try navigating around some files and folders that you are already familiar with to get comfortable with how they look through the bash shell. A few things you may want to try are: 
* Using `ls` to learn more about what's in your current directory
* Choose a specific nested folder somewhere on your computer (e.g., the location of your most recent manuscript draft) -- challenge yourself to see if you can get there via the command line
* Try moving up and then back down again in your file hierarchy all in one line" 

This way, they're actually practicing these things while they have something they can work with in front of them. 

It may also be useful to combine the "tab completion" information into the same page as the navigation commands so that when they go on to practice, we can also suggest they practice using tab completion. This would also reduce the current wall of text of tab completion information -- instead of explaining so much to them in words, we could have them experiment -- "What happens when you try to tab complete when more than one directory could match the letters you've already typed in (e.g., pressing tab after `cd L` when the current directory holds both `Letters_of_recommendation` and `Lesson_plans` subdirectories)

PR #321, opened 2023-03-15
add less command to bash 101

drelliche:
@leemc-data-ed I added the `less` command in addition to `cat`. Please let me know if there is adequate information here to prep learners for eventually using `less` with omics stuff.

Other updates include a little revamp of the "manual" section and making the highlight boxes match the new formatting.

### DIFF
--- a/bash_command_line_101/bash_command_line_101.md
+++ b/bash_command_line_101/bash_command_line_101.md
@@ -2,7 +2,7 @@
 
 author:   Nicole Feldman and Elizabeth Drellich
 email:    feldmanna@chop.edu drelliche@chop.edu
-version: 1.3.2
+version: 1.4.0
 module_template_version: 2.0.1
 language: en
 narrator: UK English Female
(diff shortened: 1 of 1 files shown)


drelliche:
@leemc-data-ed  and @franzenr I also incorporated some of the suggestions from Issue https://github.com/arcus/education_modules/issues/212 since those also apply to this first bash module.

leemc-data-ed:
@drelliche working on this now! One thing I noticed as I'm getting started is that the highlight boxes in the Lesson Preparation section still need to be updated.

bash_command_line_101/bash_command_line_101.md
```diff
@@ -374,23 +398,68 @@ This is another sentence.
...
+**Give it a try:**
 
 ```
 cat my_sentences
 ```
 
 <div class = "learn-more">
+
 The command `cat` is very powerful three-part function that allows a reader to view, combine, or create a new version of a file. In fact `cat` is a shortening of the word "con**cat**enate."
```
leemc-data-ed, line 426: I think this could be a good place to link to the manual entry for `cat`, since it does more things than we actually talk about here.

leemc-data-ed (changes requested):
I think this is great! I just suggested a couple of changes, but otherwise lgtm!

drelliche:
@leemc-data-ed Suggestions have been incorporated! Let me know if there is anything else needed before merging.

leemc-data-ed (approved):
Lgtm!

Issue #242, opened 2022-12-06
Update highlight boxes in individual modules


drelliche:
Now that the new highlight boxes symbols have been incorporated, each module will need to be checked and updated.

- Every highlight box (except for answer boxes) should have a title in the accent color as is now shown in the example. These must be included as a script that can be copied directly from the example in the `a_sample_module` examples.
- There are now more options for highlight boxes, so make sure that the correct type is used for each box (i.e. a former `learn more` box might now be a `learn-more` or a `history` or a `cool-fact` or a `behind-the-scenes` box)
- ~~Update the module_template_version to `3.0.0`~~
- Update the version. This is a [**minor** update](https://github.com/arcus/education_modules/blob/main/versioning_guidelines.md#xminorx): x.MINOR.x

We can keep track of which modules have been updated here:

- [x] bash_103_combining_commands
- [x] bash_command_line_101
- [x] bash_command_line_102
- [x] bash_conditionals_loops
- [x] data_storage_models
- [x] data_visualization_in_ggplot2
- [x] #427
- [x] #428
- [x] demystifying_geospatial_data
- [x] demystifying_python
- [x] demystifying_sql
- [x] directories_and_file_paths
- [x] git_creation_and_tracking
- [x] git_history_of_project
- [x] git_intro
- [x] git_setup_mac_and_linux
- [x] git_setup_windows
- [x] how_to_troubleshoot
- [x] learning_to_learn
- [x] omics_orientation
- [x] pandas_transformation
- [x] python_basics_writing_python_code
- [x] r_basics_introduction
- [x] r_basics_transform_data
- [x] r_basics_visualizae_data
- [x] r_missing_values
- [x] #429
- [x] reproducibility
- [x] sql_basic
- [x] #437
- [x] statistical_tests
- [x] tidy_data
- [x] using_redcap_api

franzenr:
Not sure if we're hoping to push this through soon, but if it ends up being the case that it seems like I'll be able to get an action running that yells at us whenever we try to commit a module update w/o first incrementing the version number (I haven't had a chance to play with this in a little while bc it's been deprioritized temporarily in favor of getting our recruitment/study data collection up and running in REDCap. I can try to find time for a brief feasibility/timeline estimate sooner, though) are we open to kicking the can down the road on this until I'm able to get that implemented?

[...]

PR #322, opened 2023-03-17
Redcap api add instructions for not pushing your API token to GH

### LINKED ISSUE #279: add tip about how to not push your API token to github
Suggestion for the REDCap API module, at the end where it shows how to get the R/python code from the API playground: The code generated by REDCap includes a line that saves your API token right there in your analysis file. If folks are using git/GH, they can easily forget that line is in there and version control their script with the secret included -- since truly deleting things from git and GH is so tricky/impossible, let's instead encourage them to save secrets in a separate file right from the beginning, i.e. in this module. 
That might mean an additional short section within each of the R and python sections, or perhaps it would make the most sense to have one additional section at the end covering both (mostly language-agnostic explanation/advice, and then some little snippets showing how to implement in R and python).

rosemm:
Fixes #279 
Please check my python solution in particular! I'm not sure environment variables are actually the most sensible approach here (although my googling around supports that). If there's a better way to do this, don't hesitate to suggest that.

### DIFF
--- a/using_redcap_api/using_redcap_api.md
+++ b/using_redcap_api/using_redcap_api.md
@@ -65,6 +65,7 @@ Here's what the REDCap developers [say about REDCap](https://projectredcap.org/)
 >REDCap is a secure web application for building and managing online surveys and databases. While REDCap can be used to collect virtually any type of data in any environment (including compliance with 21 CFR Part 11, FISMA, HIPAA, and GDPR), it is specifically geared to support online and offline data capture for research studies and operations. The REDCap Consortium, a vast support network of collaborators, is composed of thousands of active institutional partners in over one hundred countries who utilize and support their own individual REDCap systems.
 
 <div class = "important">
+<b style="color: rgb(var(--color-highlight));">Important note</b><br>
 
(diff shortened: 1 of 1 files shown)


using_redcap_api/using_redcap_api.md
```diff
@@ -436,6 +436,300 @@ Click on "result" in the environment pane to open a file viewer.
...
+So `redcap_api_token` and `REDCAP_API_TOKEN` could be two different environment variables on a Unix computer, but they would refer to the same thing on a Windows computer!
+
+By convention, environment variables are usually written in all caps, but you can choose to use lower case if you prefer.
+
+</div>
+
+Create the .env file
+---
+
+The `.env` file is just a text file, so you can create it in any text editor you like.
```
leemc-data-ed, line 509: It's possible it isn't really necessary, but we could potentially add a little something about text editors vs. something like Word (which some folks might think of as a text editor.

leemc-data-ed (commented):
I just had one little comment-- but it's not even totally necessary I don't think, just something to maybe consider. Otherwise lgtm!

using_redcap_api/using_redcap_api.md
rosemm: Oh, very good call! Thank you!

drelliche:
Non-content comment: @rosemm and @leemc-data-ed this might be a good moment to update all the highlight boxes, not just the ones near the information about where to store the API token. (I'm doing a quick check of issue https://github.com/arcus/education_modules/issues/242 and commenting on already open PRs)

rosemm:
Oh yes, definitely! I didn't think to look back over the other highlight boxes for this module, but I'll do that now. Good catch!

rosemm:
@leemc-data-ed would you mind taking a look again, and let me know if you think the text editor bit is clearer now?
I also updated the highlight boxes.

leemc-data-ed (approved):
Everything lgtm!

PR #323, opened 2023-03-23
Wrapper for external modules

drelliche:
This is a template for creating a wrapper for external modules.


rosemm:
I'm going to go through and update this to distinguish instructions to authors from example text by putting instructions in html comments.

rosemm:
Just dropping a quick note here about icons, since we're proposing icons for checked/unchecked items in the list of criteria for external sources. 
We use fontawesome for our icons, available via a "kit" linked with JS. It will automatically fix icons to be screenreader friendly, but we need to remember to put in the `title` attribute for any icon with semantic meaning (vs. just decorative). For the check boxes, I put in `title="Checked"` or `title="Unchecked"` for each, but open to suggestions. 
Here's details on how fontawesome implements this, for reference: https://fontawesome.com/docs/web/dig-deeper/accessibility

rosemm:
Closing this as it's superseded by https://github.com/arcus/education_modules/pull/336

PR #328, opened 2023-03-28
R practice update

rosemm:
A few little things I noticed to fix while re-reviewing the R Practice module.


leemc-data-ed (approved):
Lgtm! Thanks for taking a look!

Issue #225, opened 2022-11-02
incorrect answers/explanations in text


franzenr:
incorrect answers or contextual explanations provided in text to try on your own exercises (not quiz questions)

- [x] "Using the pipe" page: after running `tail -3 Animals.csv | sort` answer is said to be "Now the blue morpho butterfly is first, instead of the red panda because the last three lines of Animals.csv were tiger, wolf, and blue morpho." -- the last three lines of Animals.csv are wolf, blue morpho, and Koala, so the list that comes out of this command is blue morpho, koala, wolf
- [x] "chaining multiple pipes" page: "This gives us the single numeric output 4; there are 4 lines corresponding to the 4 unique animals listed in last.txt." -- the output is 5, as there are 5 unique animals in last.txr

PR #329, opened 2023-03-28
correct incorrect descriptions of output

drelliche:
This closes https://github.com/arcus/education_modules/issues/225

### DIFF
--- a/bash_103_combining_commands/bash_103_combining_commands.md
+++ b/bash_103_combining_commands/bash_103_combining_commands.md
@@ -2,8 +2,8 @@
 
 author:   Elizabeth Drellich and Nicole Feldman
 email:    drelliche@chop.edu and feldmanna@chop.edu
-version: 1.1.0
-module_template_version: 2.1.0
+version: 1.2.0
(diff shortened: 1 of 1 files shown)


franzenr (approved):
LGTM

franzenr:
WAIT

franzenr:
version number needs to be incremented ... one moment.

franzenr (changes requested):
increment version number

franzenr:
I suggested a minor version change, since it isn't as simple as just a typo, but rather some of the info was actually incorrect, but I don't thnk the module was changed enough to warrant a major. We don't cover this exact scenario in our [versioning guidelines](https://github.com/arcus/education_modules/blob/main/versioning_guidelines.md) so I'm curious what your thoughts are/if you're ok with the version number I chose @drelliche

drelliche:
I'm good with this being a "minor version change." I had been thinking of it as a typo, but since it was actually incorrect information, I agree it should be at 1.2.0

franzenr (approved):

PR #336, opened 2023-04-05
Templates update

rosemm:
So this is going to be a pretty massive PR when everything is updated, and I thought it might be helpful to start the PR itself early so folks have a chance to see (and review!) proposed changes as they become available.


rosemm:
For the proposed additions to styles.css: 

- `gratitude` and `external-resource` boxes would be (primarily) for wrapper modules. `version-update` would be if we want to use a highlight box to alert users to recent changes in the module.
- We could use `gratitude` to thank authors of original sources (this is something we might want in standard modules as well, if they rework content from e.g. carpentries)
- The `external-resource` boxes would be specifically for wrapper modules, for linking to the external resource, as a way to draw attention to the fact that we're sending them out of the module and then they should come back. This is different from `learn-more` boxes we use in standard modules. 
- `version-update` has a [pencil](https://fontawesome.com/icons/pencil?f=classic&s=solid) as its icon
- `gratitude` has a [heart](https://fontawesome.com/icons/heart?f=classic&s=solid)
- `external-resource` has an [arrow pointing out of a box](https://fontawesome.com/icons/arrow-up-right-from-square?f=classic&s=solid)
Thoughts?

_module_templates/module_macros.md
```diff
@@ -0,0 +1,328 @@
...
+
+@overview
+<div class = "overview">
+
+## Overview
+@comment
+
+**Is this module right for me?** @long_description
+
+**Estimated time to completion:** @estimated_time
```
franzenr, line 19:
```suggestion
**Estimated time to completion:** @estimated_time_in_minutes minutes
```

to avoid the issue of us writing time estimates in various ways, I'm thinking we include the text for the unit within the larger overview macro, and then update the "estimated_time" YAML field to "estimated_time_in_minutes" and then do validation in the rest of the YAML linting process to ensure that the value is an integer between 0-60.

_module_templates/standard_module_template.md
```diff
@@ -3,76 +3,76 @@
...
 language: en
 narrator: UK English Female
+
 title: Module Title
+
 comment:  This is a short, focused description of the module.
+
 long_description: This is a longer description, which should be understandable for a lay audience. It will print under "Is this module right for me?" in the overview.
+
 estimated_time: This is rough guess of how long it might take a learner to work through the module. It will print under "Estimated time to completion" in the overview
```
franzenr, line 17:
```suggestion
estimated_time_in_minutes: This is rough guess of how long it might take a learner to work through the module. It will print under "Estimated time to completion" in the overview. Valid values are any integer 1-60. 
```

in concordance with suggestion described on module_macros.md

_module_templates/wrapper_module_template.md
```diff
@@ -0,0 +1,475 @@
...
+language: en
+narrator: UK English Female
+
+title: Module Title
+
+comment:  This is a short, focused description of the module.
+
+long_description: This is a longer description, which should be understandable for a lay audience. It will print under "Is this module right for me?" in the overview.
+
+estimated_time: This is rough guess of how long it might take a learner to work through the module. It will print under "Estimated time to completion" in the overview
```
franzenr, line 17:
```suggestion
estimated_time_in_minutes: This is rough guess of how long it might take a learner to work through the module. It will print under "Estimated time to completion" in the overview. Valid values are any integer 1-60. 
```
In concordance with suggestions on module_macros.md and standard_module_template.md

_module_templates/wrapper_module_template.md
```diff
@@ -0,0 +1,475 @@
...
+
+- identify key elements
+- create a product
+- do a task
+- articulate the rationale for something
+@end
+
+
+import: https://raw.githubusercontent.com/arcus/education_modules/template_test/a_sample_module_template/module_macros.md
+-->
```
franzenr, line 42: Given the fact that a lot of these values are ones that appear in other module templates, are there any ideas for reducing the number of places that are required to be updated any time we have a change? does that seem like a worthwhile enough thing to think about given effort-value trade-offs? 

I'm thinking especially about instances where we may want to add a YAML field to all of the module types, or change one that already exists in all of them (e.g., currently with estimated_time) -- how reliable do we think we'll be able to be at ensuring they stay consistent across all the templates?

franzenr (commented):
more review to come later after rebase

_module_templates/module_macros.md
```diff
@@ -0,0 +1,328 @@
+<!--
+
+author:   DART Team
+email:    dart@chop.edu
+version:  1.0.0
+language: en
+narrator: UK English Female
+title: Module Macros
+comment:  This is placeholder module to save macros used in other modules.
```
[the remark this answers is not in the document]
rosemm: It needs to be a functional liascript doc, unfortunately.

_module_templates/wrapper_module_template.md
rosemm: I feel optimistic about keeping three templates up to date, at least for now. We've done a nice job of improving processes in an agile way for this so far -- maybe if we start to notice the templates are becoming a pain to maintain, we can optimize then.
If we did want to remove copied text across templates, I think the way to do that would be to have e.g. a bash script to generate the template files from standardized bits of text (so we'd have a txt file of the YAML items that are the same across all templates, and then we'd have files of the YAML items that are specific to each template, and we'd have a script combine them appropriately when generating the templates themselves). Right now that feels like more work to create and maintain than time/errors it would save? Maybe you have a better idea, though.

rosemm:
Trashing this PR to start a fresh one rebased from main

_module_templates/module_macros.md
franzenr: when I last looked (a few months ago, granted) all our modules were under 1 hour, just written in a variety of ways. 

I have memory of <60 mins being a hard rule, but I can't find that anywhere now (not in QA template, e.g., or in how-to.md) I also can't find any evidence of it being a soft rule (that we're supposed to shoot for it, as you said) but the fact that you also remember that makes me feel less uncertain. 

I wonder though, if one option is to restrict to under 60 mins, and then if down the line a module comes around that requires a higher limit, we can just chat about it as a team then, and adjust the hard cap to accommodate whatever we want that higher upper limit to be (perhaps 90, or whatever makes sense in the given context). 

That way, if we ever do deviate from 60 mins, we're doing so as a deliberate decision, and we have documentation of us adjusting that criteria rather than a slow creep in length.

rosemm: That sounds like a good approach to me!

rosemm:
Okie dokie, friends! This is ready for review. Because there's kind of a lot going on here, here's an executive summary.

First, I'm replacing our old `a_sample_module_template` directory and template file with the following:

- 3 template files, which are designed to be minimal so authors just open them and start writing (vs opening them, deleting a ton of stuff, and then writing)
- macro files that save standardized text used for overview, feedback, and lesson prep sections as well as scripts we use throughout (gifPreload, for example) and SQL tables. 
- Note that all the examples and instructions that used to be in the template file are now in docs.md, which has already been merged to main on #352 . This PR does include updates to docs.md, though.

I also deleted the quality assurance template as that's covered by the issue templates from #335

drelliche (approved):
LGTM, we will find typos and small changes as we start updating modules to use these templates and macros!

Issue #343, opened 2023-04-24
Broken links in Additional Resources


franzenr:
Record ID: 363 

user reports both links in additional resources section are broken.

franzenr:
@pm0kjp I realize I just assigned this to you, but also, looking at the raw file closer and trying to copy-paste the links into browser, I realized what's wrong -- the period at the end of the sentence is being interpreted as part of the link, so when a user clicks it, they're taken to the correct URL + . which of course, is not the correct URL. I can hop in and fix it really quickly.

franzenr:
I'm realizing also, that the links being formatted like this is also in violation of accessibility best practices, since individuals using a screen reader would just hear the URL read out verbatim, rather than being provided with helpful context as to what the link will take them to. I'll address that as well in my PR.

PR #344, opened 2023-04-24
Fix database normalization links

franzenr:
links weren't working correctly in liascript, as the periods at the end of the URL text were being interpreted as _part of_ the URL.

Using the "raw" link instead of a descriptive link also goes against accessibility best practices -- addressed both of these with one change.

resolves issue #343


pm0kjp (approved):
Thank you!

pm0kjp (approved):
Thank you!

PR #345, opened 2023-04-25
fix/update while loop reference

drelliche:
I moved the highlight box one page later so that I could reasonably include the content of the now defunct link (while vs for loops) in the box itself.


rosemm (approved):
LGTM!

PR #346, opened 2023-05-02
Remove education_module wrapper from schema, updates json stub-record; add yml stub-record

pakstisj:
The `stub-record.yml` file just pulls the yml front matter from `bash_command_line_101.md`. `stub-record.yml` is now valid yml. 

This involved encapsulating the @ beginning keys with double quotes as well as those multi-line values. It also shoves `@end` into the double quote values. This is still not ideal but I wanted to push this into the example to serve as a start to work with. Getting closer!

### DIFF
--- a/assets/schemas/README.md
+++ b/assets/schemas/README.md
@@ -1,0 +1,21 @@
+# On-the-fly Validation for YAML in VSCode
+
+To allow for on-the-fly validation in VSCode, you need to update the settings in your instance of VSCode or supply the correct relative path in a $schema key in the archival-request.yml file.
+
+- Install the [Red Hat YAML extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) for VSCode.
+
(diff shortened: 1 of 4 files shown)


drelliche (approved):
I'm not sure how to verify that it verifies successfully, but I look forward to learning!

drelliche (commented):
After my conversation with @rosemm , I think that the quotes that will fix the validation problem will, unfortunately, break LiaScript's ability to parse the front matter. 

However if those quotes are added only when processing the front matter into stub-record.yml for validation, it might work great? Let's dig deeper in our next meeting :)

pakstisj:
I added a README into the assets/ dir that desribces how I've set up my settings to allow yaml validation. We may be moving away from this cf. most recent previous comment and email, so we could potentially remove this file in the future, but wanted to push now in case it does turn out useful for anyone else in testing and investigation!

PR #352, opened 2023-05-05
Roseh docs

rosemm:
Add the `docs.md` file that will replace all the example text previously stored in templates. Yay!

Here it is rendered in lia, if you like: https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/roseh_docs/docs.md

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -40,7 +40,7 @@ To see how markdown works in LiaScript, look over the [LiaScript documentation o
 <b style="color: rgb(var(--color-highlight));">Behind the scenes</b><br>
 
 This document is itself a LiaScript markdown file! 
-To see what the text looks like, open the [raw version of this file]().
+To see what the text looks like, open the [raw version of this file](https://raw.githubusercontent.com/arcus/education_modules/roseh_docs/docs.md).
 
 </div>
 
--- a/docs.md
+++ b/docs.md
@@ -20,7 +20,7 @@ LiaScript is a [markdown](https://en.wikipedia.org/wiki/Markdown) dialect, so wr
 
 To see this document as an interactive LiaScript rendered version, click on the following link/badge:
 
-[![LiaScript](https://raw.githubusercontent.com/LiaScript/LiaScript/master/badges/course.svg)](https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/roseh_docs/docs.md)
+[![LiaScript](https://raw.githubusercontent.com/LiaScript/LiaScript/master/badges/course.svg)](https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/main/docs.md)
 
 </div>
 
@@ -40,7 +40,7 @@ To see how markdown works in LiaScript, look over the [LiaScript documentation o
 <b style="color: rgb(var(--color-highlight));">Behind the scenes</b><br>
 
 This document is itself a LiaScript markdown file! 
-To see what the text looks like, open the [raw version of this file](https://raw.githubusercontent.com/arcus/education_modules/roseh_docs/docs.md).
+To see what the text looks like, open the [raw version of this file](https://raw.githubusercontent.com/arcus/education_modules/main/docs.md).
 
 </div>
 
@@ -53,7 +53,7 @@ The instructions in this documentation will help you make sure your module will 
 <b style="color: rgb(var(--color-highlight));">Behind the scenes</b><br>
 
 Our QA process is very transparent. 
-You can view the [QA checklist](https://github.com/arcus/education_modules/blob/templates_update/.github/ISSUE_TEMPLATE/qa-for-standard-modules.md) that will be used to check your module at any time. 
+You can view the [QA checklist](https://github.com/arcus/education_modules/blob/main/.github/ISSUE_TEMPLATE/qa-for-standard-modules.md) that will be used to check your module at any time. 
 You can also see revisions and comments on [pull requests for other modules that have already been QAed](https://github.com/arcus/education_modules/pulls?q=is%3Apr+is%3Aclosed+label%3A%22Quality+Assurance%22) to see what kinds of adjustments were made. 
 
 </div>
--- a/docs.md
+++ b/docs.md
@@ -257,7 +257,7 @@ The learning objectives will be automatically printed twice in each module: Once
 ### `import`
 
 ```
-import: https://raw.githubusercontent.com/arcus/education_modules/main/_module_templates/macros.md
+import: https://raw.githubusercontent.com/arcus/education_modules/templates_update/_module_templates/macros.md
 ```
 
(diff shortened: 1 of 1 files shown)


docs.md
```diff
@@ -0,0 +1,1366 @@
...
+
+## Style and formatting requirements
+
+- The title should be the only level-1 header in the module document.
+- LiaScript will create a new page at each level 1, 2, or 3 header, so to avoid a page with only a header and no content, include text after each header before the next.
+- We have a standardized naming convention and directory structure:
+
+  * Folder and file names use lowercase and underscores (no dashes)
+  * Main module directory folder name is identical to the name of the module content markdown file.
+  * Images, videos, and other audio-visual assets are saved within a `media` folder within the module directory
```
drelliche, line 69: Some modules also have a `data` folder for jupyter notebooks that is separate from the `media` folder, see pandas_transform and elements_of_maps. It may be that these modules need to be updated to have the correct file structure, but it may also be useful to have both a `media` and a `data` file like elements_of_maps has.

docs.md
```diff
@@ -0,0 +1,1366 @@
...
+
+This value will already be filled in for you when you open the template file. 
+
+When you're ready to submit your module for QA, it should have a template version number that is up to date with the current module template (`template_standard.md`, `template_wrapper.md`, or `template_exercise.md`) -- if not, the module should be brought in line with any changes that have occurred to the module template before continuing submitting it for QA.
+
+### `language`
+
+```
+language: en
+```
```
drelliche, line 157: It might be helpful to add an explanatory note here that this is used if a translation of the module is ever created.

docs.md
```diff
@@ -0,0 +1,1366 @@
...
+
+### `narrator`
+
+```
+narrator: UK English Female
+```
+
+See the [LiaScript documentation on `narrator`](https://liascript.github.io/course/?https://raw.githubusercontent.com/liaScript/docs/master/README.md#189) for more details.
```
drelliche, line 165: Should we explicitly tell our users that this designates the automated voice if a user decides to listen to the module.

docs.md
```diff
@@ -0,0 +1,1366 @@
...
+ - `macros_sql_table_allergies.md` loads the allergies table
+ - `macros_sql_table_patients.md` loads the patients table
+ - `macros_sql_table_observations.md` loads the observations table
+
+Note that very small tables (just a few rows) can be constructed right in the front matter for the module that uses them. Any larger tables, or tables that need to be re-used across different modules, should be created in separate table macro modules, though.
+
+<div class = "learn-more">
+<b style="color: rgb(var(--color-highlight));">Learning connection</b><br>
+
+To learn about creating a new macro, such as a new SQL table, see our [instructions for DART macros](https://github.com/arcus/education_modules/blob/main/macros_instructions.md).
```
drelliche, line 823: This link doesn't yet exist (I think).

docs.md
```diff
@@ -0,0 +1,1366 @@
...
+### Interactive R
+
+If your lesson includes R code examples hosted in binder from [our R notebook repository](https://github.com/arcus/education_r_environment), then you'll need to load the R macros by adding the following `import` statement to the front matter of your module:
+
+```
+import: https://raw.githubusercontent.com/arcus/education_modules/templates_update/_module_templates/macros_r.md
+```
+
+You can then insert the `@lesson_prep_r` macro, which includes instructions for opening up the relevant rmd document, in your Lesson Preparation section:
+
```
drelliche, line 784: Additional front matter guidelines are being created, so we should make sure to update this section to include things like `coding_language`, `depends_on_knowledge_available_in`, etc. and make clear what is optional and when various designations are required.

docs.md
```diff
@@ -0,0 +1,1366 @@
...
+For accessibility for learners using screenreaders, if your question is more than a single markdown paragraph, you must enclose it in `<div>` tags. For example:
+
+```
+<div>
+In the following code block, what should fill in the blank?
+
+```r
+install.packages(____)
+```
+</div>
```
drelliche, line 1146: Should this `</div>` be inside the code block?

docs.md
```diff
@@ -0,0 +1,1366 @@
...
+# in this case, we can make the dots of the scatterplot disappear from any of the plots we made above
+# for example:
+ggplot(breast_cancer_data, mapping = aes(y=Glucose, x=Age)) +
+  geom_point(alpha = 0) +
+  geom_line(mapping = aes(y=model$fitted.values)) +
+  theme_bw()
+
+# this has the advantage of keeping the scales for the plot consistent
+# and it means you don't have to set the scales manually when using geom_abline
+```
```
drelliche, line 1366: Would it make sense to have a feedback form at the end of this to actually get feedback from anyone who tries to use it?
I know the feedback macro will be in the templates, and people will be unlikely to use this documentation linearly with an ending point where they want to give feedback, but we might get some good user insights if we ask for them.

docs.md
rosemm: Oh, yeah, let's add a folder for that! Is it always code/notebooks? Because maybe `data` isn't the right name, then (maybe `src`?) This would be a sensible moment to change that convention, if we want.

docs.md
rosemm: Good idea, done!

docs.md
rosemm: Good idea, done!

docs.md
rosemm: Indeed it doesn't --- I'll change it to the temporary link on the template_updates branch for clarity

docs.md
rosemm: Ah, I messed up the nesting code blocks there, good catch! Fixed now.

docs.md
rosemm: I like this idea. I think we'd want a different survey for it, though, right? We'd want slightly different questions than what we ask in the quick module feedback form. 
I've been wondering about whether we want to have a Quick General Feedback survey, for folks to submit anonymous suggestions that aren't tied to a particular module. This could maybe fit into something like that.

docs.md
rosemm: Yes, absolutely! There will be new front matter items coming from the wrapper template and macros as well (and possibly the exercise ones?). What do you think about getting this merged as the base version of the documentation, and then having updates to docs.md be included in all the other various PRs that would add new front matter fields? Rather than trying to make this version of docs.md complete before those other files are quite ready yet.

docs.md
drelliche: `src` seems like a great option for that, encompassing all sorts of files from raw data to jupyter notebooks etc.

docs.md
drelliche: I agree, let's merge what you already made, and then add to that.

docs.md
drelliche: Oooh, I really like the idea of a quick general feedback survey, like for people who have already filled out the interest form and have some general thing to tell us :)

drelliche (approved):

PR #369, opened 2023-05-10
Knowledge graph of modules

drelliche:
In order to incorporate the new metadata into the module discovery app, the app needs to live in the same branch as the new metadata.


leemc-data-ed (approved):
Lgtm!

PR #372, opened 2023-05-15
Mll docs update

leemc-data-ed:
Adding some info about exercise modules, and a little of supporting info (in quizzes and content sections). I actually decided to have a separate exercise template to avoid confusion, even though it's super similar to the standard, but that change is in templates_update branch.


rosemm (approved):
Lgtm

PR #377, opened 2023-05-25
Add level 1 header to SQL table macro files

rosemm:
Apparently liascript can't load a macro course via `import` unless it has a level 1 header after the front matter. Lesson learned!


rosemm:
Covered by #381

PR #380, opened 2023-05-25
added demystifying containers module

leemc-data-ed:
This is a pretty rough first draft of Demystifying Containers, but I think it's at the point now to get some feedback.


demystifying_containers/demystifying_containers.md
```diff
@@ -0,0 +1,165 @@
...
+
+</div>
+***
+
+## Container technology
+
+There are a variety of choices when it comes to containerization technology, but the most popular is **Docker**. [Docker's website](https://www.docker.com/) has many resources and articles to help you get started, as well as a [registry for container images, called DockerHub](https://hub.docker.com/), where you can store your own container images and make use of images already out there. 
+
+However there are other providers that you might come across, that are very similar to Docker but might have subtle differences that may make them better for specific workflows (don't worry about those differences right now though). Here are just a few:
+
```
rosemm, line 135: Are you able to provide any context about the workflows for which these other tools might be relevant? Like "If you do omics work, you might prefer Buildah, and if you do image analysis you might prefer rkt." Or something.
If not, I think it may be helpful to add a note to give folks more context for how to think about these (e.g. "These container providers all work pretty similarly. In general, you'll want to pick the container provider that seems the most widely used in your field; for most researchers that will be Docker, but if you tend to see another type of container instead, know that the principles outlined in this lesson still apply, even if the implementation might be a little different." Although I definitely just made that up, so don't say it if it's not true! ;) )

demystifying_containers/demystifying_containers.md
```diff
@@ -0,0 +1,165 @@
...
+You can also run containers on cloud platforms like [Amazon Web Services, or AWS](https://aws.amazon.com/) and [Google Cloud Platform, or GCP](https://cloud.google.com/) using various orchestration (or automation of the operational tasks that are involved in deploying and managing containers) and computing tools. You can check out [containers in AWS](https://aws.amazon.com/containers/services/) and [containers in GCP](https://cloud.google.com/containers) for more information. 
+
+### Quiz: Container technology
+
+True or false: A Docker container is the only kind of container.
+
+[( )] True
+[(X)] False 
+
```
rosemm, line 148:
```suggestion
```

Can't have this newline here or the quiz question doesn't render

demystifying_containers/demystifying_containers.md
```diff
@@ -0,0 +1,165 @@
...
+While Docker is the most popular container platform, and has many helpful resources to help you get started, it is not the only option. 
+
+</div>
+***
+
+## Additional Resources
+
+- [Another brief description of what a container is from the Docker website](https://www.docker.com/resources/what-container/)
+- [Article with more information about containers vs. virtual machines](https://www.atlassian.com/microservices/cloud-computing/containers-vs-vms) 
+
```
rosemm, line 161: I'm wondering if the "containers in the cloud" stuff you mention at the end of the last section maybe belongs here instead of there? I could see it either way.

rosemm (commented):
I think this looks really solid! I learned a lot -- it provides enough overview without getting into the weeds, which is exactly what it should do :)

rosemm:
I like your first quiz question, and I've been trying to think of a way to bring that flavor more to the forefront of the module; what do you think about adding a short sub-section (like one paragraph) called "when to NOT use a container" that just spells out that containers are only really valuable when you're doing scripted work not point and click, and only if you want to run those scripts on not-only-your-machine. And also that you should be using open-source software --- building a container for your Stata analysis (can you even do that?) isn't super helpful since someone else with the container still wouldn't be able to use it unless they also had a license for Stata. 
In a similar vein, I think you could put a little more emphasis on the value of containers for research per se, not just development. I don't think this needs a whole new subsection, you could just pop in a line (or a highlight box) saying that there have been recent calls to raise expectations for code sharing for published research and sharing just the analysis scripts themselves often doesn't work. Containers offer a much better solution for sharing your code (you could cite https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9131851/ and https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8861064/).

rosemm:
In case you were as curious as I was: Apparently you totally can use software that requires a license like Stata: https://aeadataeditor.github.io/posts/2021-11-16-docker#an-example-using-stata-licenses-and-the-easy-way But it would definitely limit the number of people who would be able to replicate your work, obvi.

demystifying_containers/demystifying_containers.md
```diff
@@ -0,0 +1,165 @@
...
+email:    leemc@chop.edu
+version:  0.0.1
+module_type: standard
+module_template_version: 3.1.0
+language: en
+narrator: UK English Female
+
+title: Demystifying Containers
+
+comment: Containers can be a useful tool for reproducible workflows and collaboration. This module aims to describe what containers are, why a researcher might want to use them, and how to get started.
```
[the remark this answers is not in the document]
leemc-data-ed: I like yours better!

demystifying_containers/demystifying_containers.md
leemc-data-ed: Oooh good call! Stupid newlines.

demystifying_containers/demystifying_containers.md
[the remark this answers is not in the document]
leemc-data-ed: Oooh yeah good catch, I meant to do that and forgot!

demystifying_containers/demystifying_containers.md
```diff
@@ -0,0 +1,165 @@
...
+title: Demystifying Containers
+
+comment: Containers can be a useful tool for reproducible workflows and collaboration. This module aims to describe what containers are, why a researcher might want to use them, and how to get started. 
+
+long_description: Writing code in multiple environments can be tricky. Collaboration can be hindered by different language versions, packages, and even operating systems. Developing code in containers can be a solution to this problem. In this module, we'll describe what containers are at a high level and discuss how they might be useful to researchers. 
+
+estimated_time_in_minutes: 20
+
+@prerequisites
+- some familiarity with programming is helpful, but not required
```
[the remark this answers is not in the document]
leemc-data-ed: Cool, done and done!

demystifying_containers/demystifying_containers.md
```diff
@@ -0,0 +1,165 @@
...
+
+Generally, to get started making your container image, you'll need a couple of things:
+
+- A set of instructions for building your container: This is usually the most difficult part of containerization. Usually you will start with a base image, then add anything else required. 
+- A list of required packages: This lets the container platform you are using know what libraries to install. This is generally in a separate file from the other instructions, so that it is easy to update. 
+
+After you have created your image, then the container platform you are using can take your instructions and build the container itself, a small computing environment that runs on your computer, but isolated from it, with only the components that it needs and nothing more. It only has access to what you have put inside it via your instructions, and once you're done, you can turn it off and it no longer exists until the next time you need it. You can even attach folders from your computer to the container, if you need to save data before the container stops running. 
+
```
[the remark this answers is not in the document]
leemc-data-ed: Agreed, I think this is worth calling out very explicitly!

demystifying_containers/demystifying_containers.md
leemc-data-ed: So it seems (at least in my brief searching around) that most folks either use docker or one of the cloud options, but I'd be uncomfortable making a blanket statement. And I don't think I'd want to try to make a statement about specific workloads (I haven't found a resource that spells it out really, and honesty a lot of the differences are pretty small and get a lot into some computer science weeds that I think might be confusing). Let me see if I can come up with something better than I have now though.

demystifying_containers/demystifying_containers.md
leemc-data-ed: I've re-worked this section, so I hope it makes more sense now (if it doesn't we can totally revisit)

demystifying_containers/demystifying_containers.md
leemc-data-ed: Okay, I've added some more stuff here (it's a little hand-wavy, but possible enough?)

leemc-data-ed:
@rosemm okay, I've address almost everything I think (I think I'm going to need to re-make the schematics though, and I won't have time to do that before I leave, but I'll get to it once I'm back). Let me know if this flows better!

leemc-data-ed:
Okay @rosemm, I've added the new diagrams, which were made by yours truly. I think that addresses everything you mentioned? Let me know if I missed anything or if something still isn't quite right!

demystifying_containers/demystifying_containers.md
```diff
@@ -69,6 +75,30 @@ Here are a few possible scenarios:
...
+- If you are only ever going to run a script on your own computer, there's no need for containerization, which is most useful for sharing and re-using code in multiple environments. 
+
+	<div class = "learn-more">
+	<b style="color: rgb(var(--color-highlight));">Learning connection</b><br>
+
+	If you want to maintain a specific environment for running a particular script on your computer (for example, if you've got multiple analyses that require different language or package versions), you can instead use a **virtual environment**. Here are some resources about [virtual environments in R](https://posit.co/blog/renv-project-environments-for-r/) and [virtual environments in Python](https://docs.python.org/3/library/venv.html).
+
+	</div>
+
+- If you use one-and-click software for analysis, containers aren't optimal for that kind of work, because the whole purpose of a container image is to lay out a reproducible set of instructions. If you want your analyses to be transparent, sharable, and reusable, consider using scripted analyses (which can be containerized)!
```
rosemm, line 98:
```suggestion
- If you use point-and-click software for analysis, containers aren't optimal for that kind of work, because the whole purpose of a container image is to lay out a reproducible set of instructions. If you want your analyses to be transparent, sharable, and reusable, consider using scripted analyses (which can be containerized)!
```

I think that's what you meant, right?

rosemm:
Gorgeous edits! I love the new clarification on container engines vs. container orchestration etc. So helpful. I think I found one little typo, but other than that this puppy is ready to merge!

demystifying_containers/demystifying_containers.md
leemc-data-ed: Oooh yes absolutely lol, good catch

rosemm (approved):
Beautiful!

PR #383, opened 2023-05-30
Remove text with survey link that's not rendering properly

rosemm:
Little fix in the print_list macro for wrapper modules. There was a little extra line of text to add our survey link, but it's not rendering properly and the survey link is available in the feedback section anyway, of course. So just taking that out.


drelliche (approved):
lgtm

PR #418, opened 2023-07-18
Joy update r basics transform

pm0kjp:
Same stuff!  Macros and metadata for the most part, with one small typo fix while I was at it.  See if I've messed anything up by looking at 
https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/joy-update-r-basics-transform/r_basics_transform_data/r_basics_transform_data.md#1


r_basics_transform_data/r_basics_transform_data.md
```diff
@@ -20,98 +22,67 @@ After completion of this module, learners will be able to:
...
+* Use the Environment tab to find a data frame and learn more about it
+* Insert a new code chunk in an R Markdown document
 
-	// Change the image to .gif when clicked and vice versa.
-	$('figure').on('click', function() {
+@end
 
-		var $this   = $(this),
-				$index  = $this.index(),
+r_file: r\_basics\_transform\_data
```
rosemm, line 35:
```suggestion
```

You had this field in there twice

r_basics_transform_data/r_basics_transform_data.md
```diff
@@ -20,98 +22,67 @@ After completion of this module, learners will be able to:
...
 
-				$img    = $this.children('img'),
-				$imgSrc = $img.attr('src'),
-				$imgAlt = $img.attr('data-alt'),
-				$imgExt = $imgAlt.split('.');
+good_first_module: true
```
rosemm, line 37: Just double checking this field --- I think Elizabeth is intending to use this for automatic generation of "pathways", and this field would be used to determine modules that can appear first in a pathway. We have only a couple modules with this set to true (reproducibility, maybe one more?). Not disagreeing, I think this could be a good way to start a pathway, but I just wanted to flag this in case it was an error.

r_basics_transform_data/r_basics_transform_data.md
```diff
@@ -20,98 +22,67 @@ After completion of this module, learners will be able to:
...
+data_task: data_wrangling
+coding_required: true
+coding_level: basic
+coding_language: r
+sequence_name: r_basics
```
rosemm, line 42:
```suggestion
sequence_name: r_basics
previous_sequential_module: r_basics_visualize_data
```

rosemm (changes requested):
Two teeny fixes in the metadata, and one question

r_basics_transform_data/r_basics_transform_data.md
pm0kjp: Oh yep I don't think this is a good first module!  That was a mistake!  Good catch!

pm0kjp, line 37:
```suggestion
good_first_module: false
```

rosemm (approved):

PR #435, opened 2023-07-21
Seaborn update

### LINKED ISSUE #428: highlight boxes for data_visualization_in_seaborn

rosemm:
Fixes #428 

Updates highlight boxes, and brings up to date with docs v1 (front matter updated, macros replace text, removed link to another module in additional resources)


pm0kjp (approved):
@rosemm I'm approving this, but if you want to make it a smidge better, I'd put something just after the first import that says something like "if you're not seeing any output after running this cell, that's expected behavior!".  Even though I do python on the regular I was waiting for it to run!  I'll let you add or not on your discretion and then merge when you see fit.

rosemm:
Good call! Thanks, @pm0kjp!

PR #440, opened 2023-07-25
Regex boundaries anchors

pm0kjp:
@drelliche This 4th module is also flags, boundaries, anchors, and greedy, because that's just how it has to be...


regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
```diff
@@ -0,0 +1,419 @@
...
+
+## Scope of This Module
+
+We consider this module to be intermediate, because it assumes basic knowledge covers more advanced regular expressions topics, without delving into the advanced topic of using regular expressions within the context of code.  
+
+Specifically, we'll address these topics:
+
+* **Anchors**, which help you define patterns that belong in the beginning and end of a string as a whole
+* **Boundaries**, which help you define patterns that belong in certain places that appear within a string
+* **Exclusions**, which allow you to look for patterns that exclude certain characters
```
drelliche, line 81: This doesn't match the table of contents. I'd be okay with taking this page out entirely, but I think if it is there it should match the order and relative importance of each topic indicated by the table of contents.

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
```diff
@@ -0,0 +1,419 @@
...
+
+## Regular Expression Flags
+
+The regular expression checker we refer to and use for this module, [Regex 101](https://www.regex101.com), has, by default, the "global" and "multiline" **flags** set, which you can see written as "gm" after the "Regular Expression" box.  Flags in regex are parameters which can change how regular expressions are interpreted.
```
drelliche, line 85:
```suggestion
**Flags** in regex are parameters which can change how regular expressions are interpreted.  The regular expression checker we refer to and use for this module, [Regex 101](https://www.regex101.com), has, by default, the "global" and "multiline" flags set, which you can see written as "gm" after the "Regular Expression" box.   
```

switching the order of these two sentences

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
```diff
@@ -0,0 +1,419 @@
...
+* **Multiline (m)**.  This flag indicates that each line should be treated as its own string, instead of all the lines being considered one string.  This is important when working with start-of-string and end-of-string **anchors**, which we'll talk about later in this module.
+* **Ignore / Insensitive (i)**.  "Ignore" indicates that pattern matching should be case-insensitive, so that differences in capitalization are ignored.
+
+If you use different "flags" than we do (for example, if you're not using the default [Regex 101](https://www.regex101.com) as your regex checker), you might see an occasional difference in how the regular expression we show works in your environment.  In the regex checker you're using, try setting your flags to "global" and "multiline", and you should get identical results to what we show here in this module.  The flags usually are shown following a forward slash, appearing just after the regular expression ends:
```
drelliche, line 93:
```suggestion
If you use different "flags" than we do (for example, if you're not using the default [Regex 101](https://www.regex101.com) as your regex checker), you might see an occasional difference in how the regular expression we show works in your environment.  In the regex checker you're using, try setting your flags to "global" and "multiline," and you should get identical results to what we show here in this module.  The flags usually are shown following a forward slash, appearing just after the regular expression ends:
```

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
```diff
@@ -0,0 +1,419 @@
...
+
+![A regular expression checker box labeled "regular expression" includes a forward slash, the phrase "insert your regular expression here", a forward slash, and the letters "g" and "m". ](media/example1.png)<!-- style = "border: 1px solid rgb(var(--color-highlight)); max-width: 400px;" --> Screenshot taken from [https://regex101.com/](https://regex101.com/)
+
+![A sample regular expression: `([A-Z])\w+` is enclosed within forward slashes.  After the closing forward slash, the letter "g" appears.](media/example2.png)<!-- style = "border: 1px solid rgb(var(--color-highlight)); max-width: 300px;" --> Screenshot taken from [https://regexr.com/](https://regexr.com/)
+
+![A regular expression checker box labeled "Regular Expression" that includes two forward slashes followed by the letter "g".](media/example3.png)<!-- style = "border: 1px solid rgb(var(--color-highlight)); max-width: 400px;" --> Screenshot taken from [https://www.regextester.com/](https://www.regextester.com/)
```
drelliche, line 99:
```suggestion
![A regular expression checker box labeled "regular expression" includes a forward slash, the phrase "insert your regular expression here," a forward slash, and the letters g and m. ](media/example1.png)<!-- style = "border: 1px solid rgb(var(--color-highlight)); max-width: 400px;" --> Screenshot taken from [https://regex101.com/](https://regex101.com/)

![A sample regular expression: `([A-Z])\w+` is enclosed within forward slashes.  After the closing forward slash, the letter "g" appears.](media/example2.png)<!-- style = "border: 1px solid rgb(var(--color-highlight)); max-width: 300px;" --> Screenshot taken from [https://regexr.com/](https://regexr.com/)

![A regular expression checker box labeled "Regular Expression" that includes two forward slashes followed by the letter g.](media/example3.png)<!-- style = "border: 1px solid rgb(var(--color-highlight)); max-width: 400px;" --> Screenshot taken from [https://www.regextester.com/](https://www.regextester.com/)
```

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
```diff
@@ -0,0 +1,419 @@
...
+</div>
+
+### Quiz: Flags
+
+Which of the following statements are true?  Check all that apply.
+
+[[ ]] Flags are only important if you're working in Perl.
+[[ ]] Flags are written as the last letter or letters of a regular expression.
+[[X]] Flags alter the rules around regular expression parsing.
+[[X]] One common flag is "global", written with a "g", which allows for multiple matches.
```
drelliche, line 117:
```suggestion
[[X]] One common flag is "global," written with a g, which allows for multiple matches.
```

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
```diff
@@ -0,0 +1,419 @@
...
+-----
+
+If we want to indicate the start of a string, we add the caret at the beginning of our regular expression, being sure to not enclose it in square brackets.  We see an example of this in the regular expression `^(?:WT|PBMC)\-\d{5,6}`, which includes a start of string character followed by a non-capturing group with two alternatives, a hyphen, and then five to six digits.
+
+Except-for Caret
+------
+
+We can also use the caret symbol in an "except for" clause in our regular expression.  For example, what if at a given position in a pattern, any character is possible except for a hash mark (#)?  We can say "except for" by using `^` as the first character within square brackets: `[^#]`.  
+
+We can add multiple characters that should not appear within the square brackets.  If any character except for  #, 5, ., and % are permitted at a given location in a string, we could express that with the regular expression list `[^#5\.%]`. For example, `[ABXYZ]` means "any of these: A, B, X, Y, or Z", while `[^ABXYZ]` means "anything that **isn't** one of these: A, B, X, Y, and Z."
```
drelliche, line 198:
```suggestion
We can add multiple characters that should not appear within the square brackets.  If any character except for  #, 5, ., and % are permitted at a given location in a string, we could express that with the regular expression list `[^#5\.%]`. For example, `[ABXYZ]` means "any of these: A, B, X, Y, or Z," while `[^ABXYZ]` means "anything that **isn't** one of these: A, B, X, Y, and Z."  
```

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
```diff
@@ -0,0 +1,419 @@
...
+------
+
+We can also use the caret symbol in an "except for" clause in our regular expression.  For example, what if at a given position in a pattern, any character is possible except for a hash mark (#)?  We can say "except for" by using `^` as the first character within square brackets: `[^#]`.  
+
+We can add multiple characters that should not appear within the square brackets.  If any character except for  #, 5, ., and % are permitted at a given location in a string, we could express that with the regular expression list `[^#5\.%]`. For example, `[ABXYZ]` means "any of these: A, B, X, Y, or Z", while `[^ABXYZ]` means "anything that **isn't** one of these: A, B, X, Y, and Z."  
+
+<div class = "important">
+<b style="color: rgb(var(--color-highlight));">Important note</b><br>
+
+To use `^` as "except for", it has to immediately follow an opening square bracket, and be followed by the characters it is indicating aren't acceptable.  You can't, for example, say "all the digits except for 7" by typing `\d^7` or `[\d^7]`.
```
drelliche, line 203: If I wanted any letter except q, would the best way to do that be [A-PR-Za-pr-z]? It might be nice to have an example here showing that you can denote "any letter except q" or something like that even though it is clunky.

Or maybe just continue the example of [0-689] or [0-68-9].

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
```diff
@@ -0,0 +1,419 @@
...
+
+**Acceptable**
+
+* stated, "I want to withdraw from the study."
+* not clear on pre-registration next steps.'
+
+**Typos**
+
+* reported she "did not receive the sample".
+* chi-square would not be robust'.
```
drelliche, line 224: I actually "laughed out loud".

And then cringed as I created my own typo...

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
```diff
@@ -0,0 +1,419 @@
...
+If you're looking for the actual symbol `^` in your text, you'll have to **escape** it by adding a backslash right before the caret: `\^`.  As a reminder, that backslash is useful for all **metacharacters** in regular expressions.  You'd do the same thing for the end-of-string character: `\$` is how you look for a literal dollar sign / money symbol.
+
+### End-of-string Anchoring 
+
+As a reminder, the symbol for end-of-string in a regular expression is `$`.  We can use this symbol to look for specific patterns at the end of strings.  For example, we could look for specific typos that tend to appear at the end of strings: periods that follow single or double quotation marks instead of being enclosed within the quotation.  Here are some examples of acceptable grammar and typos:
+
+**Acceptable**
+
+* stated, "I want to withdraw from the study."
+* not clear on pre-registration next steps.'
```
drelliche, line 219: It isn't quite right to end a sentence in `.'` since single quotes only ever appear inside double quotes (see rule 7 and 8b here: https://www.grammarbook.com/punctuation/quotes.asp)

I kind of feel bad for even knowing that, let alone looking it up and being right and sharing the link....

Actually, I think that as wonderful as this is and as honored as I am that I might have drilled this punctuation rule into your head, it is actually not a great example because it is particularly un-translatable... Even other languages using the same alphabet have different punctuation rules. I'll think about another typo I can try to suggest.

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
pm0kjp, line 203:
```suggestion
To use `^` as "except for", it has to immediately follow an opening square bracket, and be followed by the characters it is indicating aren't acceptable.  You can't, for example, say "all the digits except for 7" by typing `\d^7` or `[\d^7]`.  One way to do that would be `[012345689]`, and another would be `[0-689]`.  

In some regular expression engines, you can do some advanced practices such as creating a **union** between two sets (such as the set of all digits and the set of all symbols except for 7), but the way to do this is different in different languages and regular expressions environments, so we won't go into detail about that here.


```

pm0kjp: @drellich what do you think about the above?

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
drelliche: You did make me proud!!

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
drelliche: I think you mean the intersection of two sets (instead of union). Other than that I love this addition!

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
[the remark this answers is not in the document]
pm0kjp: @drelliche what say?  Is the above better? Or should I just delete the scope page altogether?

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
```diff
@@ -0,0 +1,419 @@
...
+
+Consider these test strings, which are movie sequel titles:
+
+* Speed 2: Cruise Control
+* Teenage Mutant Ninja Turtles II: The Secret Of The Ooze
+* Tomb Raider: The Cradle Of Life
+
+You're helping a colleague who tried to write regex that would capture two groups for each movie: first, the part of the title before the colon, and second, the part of the title after the colon. 
+
+They used this regular expression, which uses capturing groups:  `([A-Za-z0-9\s]+):([A-Za-z0-9\s]+)`. [As we can see in Regex 101](https://regex101.com/r/3kyzIp/1), this did not segment the movie titles exactly right.  The regular expression matches the pattern twice, not three times.  It gets the first capturing group correct.  But for the second capturing group, the parser takes the second part of the first movie and keeps reading well into the second movie line, stopping only when it gets to a colon.
```
drelliche, line 256: Question about this example: the screenshot shows that the `g` and `m` flags are on, which initially made me think that Regex101 would treat each line as its own separate string, so I was confused by this behavior. But looking more closely at the Regex101 documentation I see that the `m` flag actually means that `^` and `$` match the beginning and ends of lines, not that each line is its own string.  This leads me to things:

1. Correct the phrasing describing the `m` flag on the flags page
2. Move the flags section here so that you can then give one more example with anchors: `^([A-Za-z0-9\s]+):([A-Za-z0-9\s]+)$`

Thoughts?

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
drelliche: Oh, this is now the first introduction of the word "engine" which might benefit from a definition/disambiguation to learners new to regex.

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
```diff
@@ -0,0 +1,419 @@
...
+
+**Flags** in regex are parameters which can change how regular expressions are interpreted.  The regular expression checker we refer to and use for this module, [Regex 101](https://www.regex101.com), has, by default, the "global" and "multiline" flags set, which you can see written as "gm" after the "Regular Expression" box.   
+
+Flags aren't written within a regular expression, but are set **outside** of it, either by clicking on something (it might be the word "flag" or an image of a flag) or by typing a letter (like "g" or "m") or letters into your code, separate from the definition of a regular expression. Three frequently used flags are:
+
+* **Global (g)**.  This indicates that the entire string given for matching against the pattern should be checked.  Without the "g" or "global" flag, only the first match to the pattern will be found, and the regex parser will stop looking for additional matches.
+* **Multiline (m)**.  This flag indicates that each line should be treated as its own string, instead of all the lines being considered one string.  This is important when working with start-of-string and end-of-string **anchors**, which we'll talk about later in this module.
```
pm0kjp, line 90:
```suggestion
* **Multiline (m)**.  This flag indicates that each line should be treated as as if it were its own string.  This is important when working with start-of-string and end-of-string **anchors**, which we'll talk about later in this module.  When the multiline flag is turned on, it means that the anchors that typically only refer to the beginning and end of a string will also be used to the beginning and end of each line. 
```
@drelliche this is regarding the multiline movie parsing example further on.

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
pm0kjp: @drelliche Can we just pretend this is a reasonable thing to search for?  I could also just search for the double quote followed by a period, but I like to complicate things by giving an example where there are two typos we're looking for.

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
pm0kjp: @drelliche actually I might leave flags where it belongs and just remove this scenario portion altogether, it might be confusing and un-helpful to people.  What do you think?

pm0kjp:
@drelliche I updated with some comments -- mostly that I think removing the movies example might be wise... thoughts?

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
```diff
@@ -0,0 +1,425 @@
...
+- Use boundaries in regular expressions
+
+@end
+good_first_module: false
+coding_required: true
+coding_level: intermediate
+sequence_name: regex
+previous_sequential_module: regular_expressions_groups
+@sets_you_up_for
+- regular_expressions_lookaheads
```
drelliche, line 39:
```suggestion

```
I don't think these modules actually depend on/set you up for each other, even though they are sequential in the sequence, which is recorded elsewhere in the metadata.

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
```diff
@@ -0,0 +1,425 @@
...
+
+</div>
+
+Literal Caret
+---
+
+If you're looking for the actual symbol `^` in your text, you'll have to **escape** it by adding a backslash right before the caret: `\^`.  As a reminder, that backslash is useful for all **metacharacters** in regular expressions.  You'd do the same thing for the end-of-string character: `\$` is how you look for a literal dollar sign / money symbol.
+
+### End-of-string Anchoring 
+
```
drelliche, line 219: Ok, here is a proposed update that will satisfy my grammar prescriptivism for published materials (I try to be a descriptivist with language but punctuation is more of struggle).

```
As a reminder, the symbol for end-of-string in a regular expression is `$`.  We can use this symbol to look for specific patterns at the end of strings.  For example, we could look for specific typos that tend to appear at the end of strings: periods that follow something other than a letter or number.  Here are some examples of acceptable grammar and typos:

**Acceptable**

* This is with respect to record 117.
* Participants were not clear on pre-registration next steps.


**Typos**

* Reported she "did not receive the sample".
* A chi-squared test would not be robust . 

The pattern we're looking for is "anything that is not a letter or digit, followed by a period," or `[^\w]\.$`.  If you want to see this in action, you can [try it on Regex 101](https://regex101.com/r/4cZn0e/1).
```

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
drelliche: I agree, let's take this scenario out entirely and save it for a future regex practice module.

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
```diff
@@ -0,0 +1,425 @@
...
+<div class = "behind-the-scenes">
+<b style="color: rgb(var(--color-highlight));">Behind the scenes</b><br>
+
+There are other defined boundaries, such as the "not a word" boundary (`\B`), which isn't supported in every engine.  For now, the word boundary will probably be enough for most of the regular expressions you need to write.
+
+</div>
+
+## Quiz: Anchors and Boundaries
+
+What's the anchor character that indicates the "start of a string"?
```
drelliche, line 356:
```suggestion
What's the anchor character that indicates the **start of a string**?
```

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
```diff
@@ -0,0 +1,425 @@
...
+</script>
+***
+<div class = "answer">
+
+The anchor for the start of a string is the caret, or `^`, symbol.
+
+</div>
+***
+
+What's the anchor character that indicates the "end of a string"?
```
drelliche, line 371:
```suggestion
What's the anchor character that indicates the **end of a string**?
```

drelliche (commented):
@pm0kjp I agree with taking out the movies scenario (and saving it for later use). I also proposed a new example for the "grammar" search example, let me know if it works for you.

Also made a few more tiny tweaks but I think it is extremely close!

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
pm0kjp: I like it!

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
pm0kjp: agree!  adding to a batch for commit

regular_expressions_boundaries_anchors/regular_expressions_boundaries_anchors.md
```diff
@@ -0,0 +1,378 @@
...
+* stated, "I want to withdraw from the study."
+* not clear on pre-registration next steps.'
+
+**Typos**
+
+* reported she "did not receive the sample".
+* chi-square would not be robust'. 
+
+The pattern we're looking for is "single or double quotation followed by a period and the end of the string," or `['"]\.$`.  If you want to see this in action, you can [try it on Regex 101](https://regex101.com/r/9eMUve/1).
+
```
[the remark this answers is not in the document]
pm0kjp: This is your suggestion, but added as a committable suggestion!

pm0kjp:
@drelliche I added your superb example, TYSM for doing it in regex101 as well!!!  I think we're done?

drelliche (approved):
LGTM! Excellent regex series!

Issue #367, opened 2023-05-10
Talk about brackets syntax?


pm0kjp:
it is still confusing to me about when to use regular brackets, and when to use square ones. The last line in the for look used square brackets --- I would have never thought about it on my own. So, some clarification on when to use square brackets would be nice.

See REDCap Record 188 for more context.

leemc-data-ed:
Handled by #461

Issue #366, opened 2023-05-10
talk about when to use methods?


pm0kjp:
"I couldn't get the for loop's first line right, until I looked at the answer and found out there needs a '.items' method there; reflecting on this issue, it would be nice during the module to give a couple of more examples on when to use a method."

See REDCap record 188 for additional context!

leemc-data-ed:
Addressed in #459

Issue #362, opened 2023-05-05
while --> which typo


pm0kjp:
On Quiz: Lists, the first question is "While of the following" -> should be "which of the following".

See record 242 in REDCap

leemc-data-ed:
Addressed in #461

PR #451, opened 2023-07-31
Python basics edits

leemc-data-ed:
I've tried to explain some more about methods vs. functions and when to used subsetting notation, addressing feedback captured in #366 and #367. Also corrected a typo, addressing #362.

### DIFF
--- a/python_basics_writing_python_code/python_basics_writing_python_code.md
+++ b/python_basics_writing_python_code/python_basics_writing_python_code.md
@@ -113,7 +113,7 @@ There are many [built-in functions in Python](https://docs.python.org/3/library/
 Methods
 ======
 
-**Methods** in Python are very similar to functions, but are not precisely the same. The exact differences are not important to us right now, except that the syntax is a little different. Methods cannot be called by themselves; they are always called on an object, with the syntax `object.method()`.
+**Methods** in Python are very similar to functions, but are not precisely the same. The exact differences are not important to us right now, except that the syntax is a little different. Methods cannot be called by themselves; they are always associated with an object, with the syntax `object.method()`.
 
 Let's look at an example and try applying the `lower()` method:
 
@@ -128,7 +128,7 @@ print("Hello World!".lower())
 </lia-keep>
 </div>
 
-As you might have suspected, the `.lower()` method made all of the letters of the phrase "Hello World!" lowercase. The `.lower()` method is a **string method**, or a method that must be called on a string, of which "Hello World!" is an example. (There is also a string method called `.upper()`, can you guess what it does?) The Python documentation has [more information about string methods](https://docs.python.org/2.5/lib/string-methods.html).
+As you might have suspected, the `.lower()` method made all of the letters of the phrase "Hello World!" lowercase. The `.lower()` method is a **string method**, or a method that *must* be called on a string object, of which "Hello World!" is an example. (There is also a string method called `.upper()`, can you guess what it does?) The Python documentation has [more information about string methods](https://docs.python.org/2.5/lib/string-methods.html). Other kinds of objects in Python (like lists and dictionaries, for example, which we'll discuss later in this module) have their own associated methods.
 
 Let's talk more about strings and other data types next.
 
@@ -156,6 +156,13 @@ Notice that in the above example, we have passed the `type("Hello World!")` func
 
 </div>
 
+<div class = "care">
+<b style="color: rgb(var(--color-highlight));">A little encouragement...</b><br>
+
+You may be wondering exactly when you should use functions vs. methods, and that can be a tricky question to answer! It depends on whether the thing you want to do was written as a function or a method, and even folks with a lot of Python experience can forget and need to look things up! A good approach is to use your favorite search engine to search for the task you want to do combined with "Python". For example, if you want to figure out how to make a string completely lower case, you might search "make string lower case in Python". That will lead you to the string method `.lower()`. 
+
+</div>
+
 ### Quiz: Functions and methods
 
 Which of the following are valid examples of Python code? Select all that apply.
@@ -167,7 +174,7 @@ Which of the following are valid examples of Python code? Select all that apply.
 ***
 <div class = "answer">
 
-These are all examples of valid Python code! `len("Python is awesome!")` returns the length of the argument `"Python is awesome!"` (which is 18-- spaces and punctuation count!) and prints that output to the screen. `"python".upper()` makes the string "python" all uppercase. `type("4.3")` might be a little tricky, since 4.3 is a decimal number (or **float**), but it's also in quotation marks, which means that "4.3" is a string in this case! `type(True)` might also be tricky, since at first it looks like a string that doesn't have the required quotation marks. However, `True` and `False` are special in Python, and have the type `boolean`.  
+These are all examples of valid Python code! `len("Python is awesome!")` uses the `len()` function to return the length of the argument `"Python is awesome!"` (which is 18-- spaces and punctuation count!) and print that output to the screen. `"python".upper()` uses the string method `.upper()` to make the string "python" all uppercase. `type("4.3")` might be a little tricky, since 4.3 is a decimal number (or **float**), but it's also in quotation marks, which means that "4.3" is a string in this case! `type(True)` might also be tricky, since at first it looks like a string that doesn't have the required quotation marks. However, `True` and `False` are special in Python, and have the type `boolean`.  
 
 </div>
 ***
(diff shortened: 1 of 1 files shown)


python_basics_writing_python_code/python_basics_writing_python_code.md
```diff
@@ -224,38 +156,59 @@ Notice that in the above example, we have passed the `type("Hello World!")` func
 
 </div>
 
+<div class = "care">
+<b style="color: rgb(var(--color-highlight));">A little encouragement...</b><br>
+
+You may be wondering exactly when you should use functions vs. methods, and that can be a tricky question to answer! It depends on whether the thing you want to do was written as a function or a method, and even folks with a lot of Python experience can forget and need to look things up! A good approach is to use your favorite search engine to search for the task you want to do combined with "Python". For example, if you want to figure out how to make a string completely lower case, you might search "make string lower case in Python". That will lead you to the string method `.lower()`.
```
drelliche, line 162:
```suggestion
You may be wondering exactly when you should use functions vs. methods, and that can be a tricky question to answer! It depends on whether the thing you want to do was written as a function or a method, and even folks with a lot of Python experience can forget and need to look things up! A good approach is to use your favorite search engine to search for the task you want to do combined with "Python." For example, if you want to figure out how to make a string completely lower case, you might search "make string lower case in Python." That will lead you to the string method `.lower()`. 
```

python_basics_writing_python_code/python_basics_writing_python_code.md
```diff
@@ -196,7 +128,7 @@ print("Hello World!".lower())
 </lia-keep>
 </div>
 
-As you might have suspected, the `.lower()` method made all of the letters of the phrase "Hello World!" lowercase. The `.lower()` method is a **string method**, or a method that must be called on a string, of which "Hello World!" is an example. (There is also a string method called `.upper()`, can you guess what it does?) The Python documentation has [more information about string methods](https://docs.python.org/2.5/lib/string-methods.html).
+As you might have suspected, the `.lower()` method made all of the letters of the phrase "Hello World!" lowercase. The `.lower()` method is a **string method**, or a method that *must* be called on a string object, of which "Hello World!" is an example. (There is also a string method called `.upper()`, can you guess what it does?) The Python documentation has [more information about string methods](https://docs.python.org/2.5/lib/string-methods.html). Other kinds of objects in Python (like lists and dictionaries, for example, which we'll discuss later in this module) have their own associated methods.
```
drelliche, line 132: Right now the first time a learner would be seeing a method taking a parameter is in the quiz for this section. I really like `.endswith()` question and don't think you should use a different method, but maybe you could add an explanation here "functions take arguments but methods also take parameters." 

I suppose it could also go in the answer to the quiz question, but I think this might be a better spot.

python_basics_writing_python_code/python_basics_writing_python_code.md
```diff
@@ -808,7 +761,7 @@ Another kind of collection object in Python is a **dictionary**. Dictionaries ar
 
 * They do not allow duplicate key-value pairs.
 
-* Dictionaries use curly brackets { }.
+* Dictionaries use curly brackets `{ }` (unlike lists, which use square brackets `[ ]`).
```
drelliche, line 764:
```suggestion
* Dictionaries use curly brackets `{ }` (unlike lists, which use square brackets `[ ]`). These curly brackets can also be called **braces**.
```

python_basics_writing_python_code/python_basics_writing_python_code.md
```diff
@@ -892,9 +845,9 @@ print(contact_info)
     <script type="text/x-sage">
 
     contact_info_copy = contact_info.copy()
-contact_info_copy.pop("telephone")
-print(contact_info)
-print(contact_info_copy)
+    contact_info_copy.pop("telephone")
+    print(contact_info)
+    print(contact_info_copy)
```
drelliche, line 850:
```suggestion
contact_info_copy.pop("telephone")
print(contact_info)
print(contact_info_copy)
```
python caring about indentation again

python_basics_writing_python_code/python_basics_writing_python_code.md
```diff
@@ -903,35 +856,72 @@ print(contact_info_copy)
...
+* There isn't a method to add items to a dictionary, but it can be done! You can give the dictionary a new key and assign a value to it (notice that we're using square bracket notation again, like we did for lists). This is another example of a permanent change to our dictionary.
 
     @sage
     <div class="python_link">
     <lia-keep>
     <script type="text/x-sage">
 
     contact_info["work_email"] = "myname@company.com"
-print(contact_info)
+    print(contact_info)
```
drelliche, line 867:
```suggestion
print(contact_info)
```

python_basics_writing_python_code/python_basics_writing_python_code.md
```diff
@@ -903,35 +856,72 @@ print(contact_info_copy)
...
+    To return the keys of a dictionary, you use the `.keys()` dictionary method.  
+
+    </div>
+    ***
+
+3. How would you add the the country of Angola and its capital city of Luanda to the dictionary below?
+
+    `capital_cities = {"Afghanistan" : "Kabul", "Albania" : "Tirana", "Algeria" : "Algiers", "Andorra" : "Andorra la Vella"}`
+
+    [[capital_cities["Angola"] = "Luanda"]]
```
drelliche, line 917:
```suggestion
    [[capital_cities["Angola"] = "Luanda"]]
    <script>
  let input = "@input".replace(/\s+/g, '');
  input == "capital_cities[\"Angola\"]=\"Luanda\"";
</script>
```

I think this will work to allow for a difference in the number of spaced (e.g. I tried with no spaces around the `=`)

python_basics_writing_python_code/python_basics_writing_python_code.md
```diff
@@ -903,35 +856,72 @@ print(contact_info_copy)
...
+    While **values** can be repeated in a dictionary, each **key** must be unique. However, you can have multiple values for a key by putting those values in a list.
+
+    </div>
+    ***
+
+2. How would you access all of the counties (the "keys") in the dictionary below? 
+
+    `capital_cities = {"Afghanistan" : "Kabul", "Albania" : "Tirana", "Algeria" : "Algiers", "Andorra" : "Andorra la Vella"}`
+
+    [[capital_cities.keys() ]]
```
drelliche, line 904:
```suggestion
    [[capital_cities.keys() ]]
    <script>
  let input = "@input".trim();
  input == "capital_cities.keys()";
</script>
```

This might be unnecessary, but I just did something similar for the other question.

drelliche (commented):
@leemc-data-ed It looks pretty good, yay metadata and macros! There are some bugs, I think I got them all, but please double check (especially the one removing white space from the quiz question input in Dictionaries quiz.

One content suggestion that is to mention parameters along with methods, especially since the quiz question uses a method with a parameter now.

I'm somewhat hesitant to suggest this right before launch, and "we are not doing this right now" is a totally valid response, but I think this would be a great module to split into smaller parts. Specifically, I think the loops and conditional statements sections could just be cut out and made into their own module. I also think the Challenge Problem might be nice as its own standalone exercise module.

Let me know what you think, I'm happy to approve it as the one module it is once the bugs are fixed.

python_basics_writing_python_code/python_basics_writing_python_code.md
leemc-data-ed: Argh yep yep.

python_basics_writing_python_code/python_basics_writing_python_code.md
leemc-data-ed: Oooh, good call!

python_basics_writing_python_code/python_basics_writing_python_code.md
leemc-data-ed: Oooh yeah that's a good catch, I'll add something here.

leemc-data-ed: Okay, I did that thing!

leemc-data-ed:
Oooh, that's a thought! That could be a good idea actually... Though the challenge problem/practice in particular might require some massaging. Let's circle back tomorrow and chat about whether it's feasible before this wave or if we save it for next time.

drelliche:
This PR has been subsumed by https://github.com/arcus/education_modules/pull/466, https://github.com/arcus/education_modules/pull/461, https://github.com/arcus/education_modules/pull/459 and one still in progress to make the exercise its own module.

Issue #160, opened 2022-07-08
add link to info about installing python modules


rosemm:
on page #seaborn-in-python, add highlight box to direct learners to intro instructions for using pypi and conda for installs

PR #472, opened 2023-08-04
Add instructions for installing packages

leemc-data-ed:
Not ideal, but maybe will work for now? No pressure if not, I think we'll be okay if this doesn't make it in right now.


rosemm (approved):
Awesome, thank you!

rosemm:
Fixes #160

PR #475, opened 2023-08-04
Fix incorrect links

franzenr:
* fix links to boundaries + anchors regex module within lookaheads module
* fix link to data viz in OSS in ggplot2 module


drelliche (approved):
lgtm

PR #478, opened 2023-08-07
fix make_survey_URL macro

### LINKED ISSUE #477: feedback macro does not pipe in version correctly
@make_survey_url incorrectly assigns version number to parameter `version_number`, however, REDCap Quick Module Feedback form field is simply `version`.

franzenr:
make_survey_URL was assigning the version to fill a REDCap field that didn't exist, `version_number`. Updated to correctly set it to `version` which is the name of the field in REDCap.

this (hopefully) resolves #477

### DIFF
--- a/_module_templates/macros.md
+++ b/_module_templates/macros.md
@@ -13,7 +13,9 @@ comment:  This is placeholder module to save macros used in other modules.
 
 Previous versions: 
 
-- [1.0.0](https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/e983922162e6fbf971c03dc96052f68713cc72af/_module_templates/macros.md#1): Initial version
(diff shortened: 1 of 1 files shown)


_module_templates/macros.md
```diff
@@ -2,8 +2,8 @@
 
 author:   DART Team
 email:    dart@chop.edu
-version:  1.1.0
-current_version_description: Add current_version_description and version_history metadata
+version:  2.1.0
```
rosemm, line 5:
```suggestion
version:  1.2.0
```
I think this is probably a minor update

rosemm (approved):
This is great, thank you!
I copied in the previous version info to version_history, and bumped it down from a major revision to a minor one (although I agree that's fuzzy for macros).

PR #522, opened 2023-08-15
Git rstudio patch

### LINKED ISSUE #503: need git installed if you'll be working locally?
record number: 556

Description of issue: If using RStudio on desktop to load the project files, you also need Git installed. But this is never mentioned in the module and can be confusing.

This would be a change to the `lesson_prep_r` macro.

rosemm:
Fixes #503 
Although heads up: I don't have access to a windows machine, so I haven't tested R/RStudio/Git install instructions for Windows. I did link to happy git with r, though, which is an excellent resource for anyone facing install problems, so hopefully that helps.

### DIFF
--- a/_module_templates/macros_r.md
+++ b/_module_templates/macros_r.md
@@ -2,8 +2,8 @@
 
 author:   DART Team
 email:    dart@chop.edu
-version:  1.1.0
-current_version_description: Add current_version_description and version_history metadata
+version:  1.2.0
+current_version_description: Add instructions for git install (windows users)
 language: en
 narrator: UK English Female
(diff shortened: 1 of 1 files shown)


leemc-data-ed (commented):
I think this looks great! One suggestion, which you are free to take or leave, is that it might be useful for folks to know they can use `git --version` if they aren't sure if git is already installed, which will either give you the version or an error.

rosemm:
@leemc-data-ed Great idea, thank you! I put something in; does that look good to you?

leemc-data-ed (approved):
Lgtm!

PR #524, opened 2023-08-15
Add missing front matter fields to templates

### LINKED ISSUE #517: module templates need added front matter
If my understanding of the [docs](https://github.com/arcus/education_modules/blob/main/docs.md) is correct, then we need to add: 
* good_first_module
* sets_you_up_for
* depends_on_knowledge_available_in

to the front matter that is included in the [templates](https://github.com/arcus/education_modules/tree/main/_module_templates) for all 3 module types. 

Presumably all exercise module types will need to have both of the following (and therefore it should be added to the front matter in the templates. 
* coding_level 
* coding_language

rosemm:
Fixes #517 

Note: We're not incrementing versions for templates, since those version fields are intended to be filled in for whatever module the author is using the template for. We might want to add a way to version templates, but for now they stay at 0.0.0

### DIFF
--- a/_module_templates/template_exercise.md
+++ b/_module_templates/template_exercise.md
@@ -32,6 +32,9 @@ After completion of this module, learners will be able to:
 @end
 
 good_first_module: false
+coding_required: true
+coding_level: 
+coding_language: 
 
 @sets_you_up_for
 
--- a/_module_templates/template_exercise.md
+++ b/_module_templates/template_exercise.md
@@ -31,7 +31,7 @@ After completion of this module, learners will be able to:
 - articulate the rationale for something
 @end
 
-good_first_module: false
+good_first_module: 
 coding_required: true
 coding_level: 
 coding_language: 
--- a/_module_templates/template_exercise.md
+++ b/_module_templates/template_exercise.md
@@ -31,6 +31,16 @@ After completion of this module, learners will be able to:
 - articulate the rationale for something
 @end
 
+good_first_module: false
+
+@sets_you_up_for
+
+@end
+
+@depends_on_knowledge_available_in
+
+@end
+
 @version_history 
 
 Previous versions: 
--- a/_module_templates/template_standard.md
+++ b/_module_templates/template_standard.md
@@ -31,7 +31,7 @@ After completion of this module, learners will be able to:
 - articulate the rationale for something
 @end
 
-good_first_module: false
+good_first_module: 
 
 @sets_you_up_for
 
--- a/_module_templates/template_standard.md
+++ b/_module_templates/template_standard.md
@@ -31,6 +31,16 @@ After completion of this module, learners will be able to:
 - articulate the rationale for something
 @end
 
+good_first_module: false
+
+@sets_you_up_for
+
+@end
+
+@depends_on_knowledge_available_in
+
+@end
+
 @version_history 
 
 Previous versions: 
--- a/_module_templates/template_wrapper.md
+++ b/_module_templates/template_wrapper.md
@@ -31,7 +31,7 @@ After completion of this module, learners will be able to:
 - articulate the rationale for something
 @end
 
-good_first_module: false
+good_first_module: 
 
 @sets_you_up_for
 
--- a/_module_templates/template_wrapper.md
+++ b/_module_templates/template_wrapper.md
@@ -31,6 +31,16 @@ After completion of this module, learners will be able to:
 - articulate the rationale for something
 @end
 
+good_first_module: false
+
+@sets_you_up_for
+
+@end
+
+@depends_on_knowledge_available_in
+
+@end
+
 @version_history 
 
 Previous versions: 


franzenr:
is there a reason that good_first_module is default listed as false for all templates instead of being blank? @rosemm

rosemm:
Yeah, we could leave it blank if you prefer. It should be false for almost all modules, though --- Elizabeth is using that as a field to pick modules that should start a pathway, and she wants to be quite selective about that since it sets the tone for the whole learner experience. So my thought was we should be quite intentional about setting good_first_module to true

franzenr:
Hmm that's a good point  -- I think I'd still prefer to leave it as blank, so (a) module authors still remember that it _is_ something that can be deliberately specified and (b) the code for extracting the fields to check for, based on the templates, can be much simpler. That being said, if it's really important to default it to false (e.g., perhaps our process for deciding a module is a good_first_module will involve periodically identifying modules to update to be `true` for this field as a group, rather than people deciding on the fly during QA) I'm happy to accommodate, and it wouldn't be that much of a bear to implement -- just a little messier. 


@drelliche can you provide some insight into how you're thinking we will decide if a module is a good_first_module or not?

rosemm:
If it complicates the QA checks, then let's leave it blank! I'll update now.

_module_templates/template_exercise.md
```diff
@@ -31,6 +31,19 @@ After completion of this module, learners will be able to:
 - articulate the rationale for something
 @end
 
+good_first_module: false
```
rosemm, line 34:
```suggestion
good_first_module: 
```

_module_templates/template_standard.md
```diff
@@ -31,6 +31,16 @@ After completion of this module, learners will be able to:
 - articulate the rationale for something
 @end
 
+good_first_module: false
```
rosemm, line 34:
```suggestion
good_first_module: 
```

franzenr:
oh wait, I forgot to mention that I think it probably makes complete sense to have it auto default to false for exercise modules! I can't imagine an instance where an exercise module would be considered a good first, unless I'm misunderstanding how we envision that category!!!

franzenr (approved):
LGTM

drelliche:
I agree that it makes sense to default to false for exercise modules. 

As to when `good_first_module` would be true, the idea is that if someone had never seen any of our modules before and wanted an idea of what they could be like, this would be a place we would want them to start at. For the purpose of module discovery, these are the first modules displayed on the page, with a kind of "Want to see what we are about? Try one of these modules!" message.

@franzenr does that make sense? Is that well defined? Or do we need some more specific guidelines? I'm seeing how this could be a thing that a contributing author might not be able to determine for themself about their own module, but I think as a team we will likely have fairly consistent ideas about what modules represent us well.

franzenr:
More than the guidelines themselves, I'm curious when you're envisioning this taking place @drelliche. Should this be getting determined by a module author + QA pair? Should the field always be "false" on first merge in, until we have a chance to convene as as team and decide which new modules we want to be considered first modules? etc f


Also, for transparency, Rose H and I had some back in forth in messages about whether or not to have it default to false specifically for exercise modules, and at least in this PR, wound up having it blank for all of the modules. 

Definitely my fault for moving the communication to messages -- I sent one to her really quickly to make sure she didn't merge in without seeing the comment I left on here (forgetting that I still needed to give the approving review lol) and we wound up chatting it out there.

franzenr:
I'm copying in the messages we exchanged below, for transparency, record keeping, etc. 

If you arrive at a different conclusion than us, I think that's fine, and we can always update to be default false for exercise in a new PR. 

Importantly, it won't be terrible to have it default as "false" specifically for exercise modules if we think it will be false for ALL exercise modules. I moreso just didn't want to have to deal with the template for a given module type saying False, but it not needing to _always_ be false. (this is the same reason I don't mind having `coding_required: true` for exercise but it's blank for other templates. It will need to always be true for exercise type. 

(These messages were all sent very quickly, so they might read awkwardly since we overlap each others thoughts) 

F: "interrupting you super quickly
because I just threw another comment on bc I realized I forgot to put all of my in-my-brain words into my in-my-comment words" 
H: Does having it default to false on exercise modules not complicate QA?
F: nope, QA checks are module type specific. Depending on the specified module type, it will check against a different set of required fields. and in this case, I’m assuming it will _always_ be false for all exercise modules forever
H: Because I think there’s very slim value in setting it as a default in the template --- looking to make sure that value is sensible will still definitely be part of QA no matter what
F: true, fair point! I was thinking it might make it one fewer step, but also the lift for an author to write “false” isn’t that big so maybe that’s silly of me lol
H: So basically, I’m saying I don’t think it matters much one way or the other from the author experience re templates. We should set it up to make the automation as smooth as possible, imo
F: ok, that’s fine by me then! I thought maybe you were initially thinking there was lots of value in setting a default
H: so i’m actually not sure --- are you saying leaving it blank for all three templates is fine? or did you want me to put it back in for exercise?
F: yes lets just leave it blank for all three then sorry

PR #535, opened 2023-08-17
Docs v2 update

### LINKED ISSUE #523: make `coding_required` a mandatory front matter field
hmm, looking over things again -- should we also be requiring `coding required:` as a field for all modules, regardless of type? (where the field entry is either yes or no?) Currently this only shows up as a necessary field in the when using the front matter checklist once "learner will need to code to meet the learning objectives" is selected.

If this isn't how that field is currently envisioned to function, can I petition that we make it that way? This will make it MUCH easier to automatically check that the required `coding_level` and `coding_language` are present for the modules that require coding.

_Originally posted by @franzenr in https://github.com/arcus/education_modules/issues/517#issuecomment-1678152076_

rosemm:
Updating docs, templates, and github issue to make `coding_required` a mandatory front matter field for all modules
Fixes #523

### DIFF
--- a/.github/ISSUE_TEMPLATE/docs_update_checklist_v1.md
+++ b/.github/ISSUE_TEMPLATE/docs_update_checklist_v1.md
@@ -1,5 +1,5 @@
 ---
-name: Checklist for updating modules
+name: Checklist for updating modules to v1
 about: Checklist for updating a module to comply with docs.md v1
 title: Docs v1 Update Checklist
 assignees: ''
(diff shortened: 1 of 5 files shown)


drelliche (commented):
I think the table on the "How to use the front matter" page needs to be updated. It looks like this is line 174 in docs.md. Everything else looks good to me!

rosemm:
Good catch, thank you @drelliche !! I think this is fixed now.

drelliche (approved):
lgtm!

PR #541, opened 2023-08-17
537 docs v1 update bash combine commands

drelliche:
This PR updates the front matter to docs v1, and also introduces a new macro for the bash_basics sequence lesson prep.

Line 60 in bash_command_line_103.md can be removed before merging, but doesn't have to be, it just links to the current location of the lesson prep macro.

### DIFF
--- a/_module_templates/macros_bash.md
+++ b/_module_templates/macros_bash.md
@@ -21,7 +21,7 @@ No Previous Versions
 You will get the most out of this lesson if you follow along with the examples and try out the commands. In order to do that you need to have a bash shell open on your computer. Please follow the instructions appropriate for the computer you are using.
 
 **Open a bash shell.**
-If you are using a computer with running iOS (i.e. a Mac) you can use the **Terminal** program. If you are on a computer using Windows, open either **WLS** (Windows Linux Subsytem) or **Git Bash**. If you don't have these programs there are instructions for how to download and set them up in the [Bash / Command Line 101](https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/main/bash_command_line_101/bash_command_line_101.md) module.
+If you are using a computer with running a Unix operating system (i.e. a Mac or Linux computer) you can use the **Terminal** program. If you are on a computer using Windows, open either **WLS** (Windows Linux Subsytem) or **Git Bash**. If you don't have these programs there are instructions for how to download and set them up in the [Bash / Command Line 101](https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/main/bash_command_line_101/bash_command_line_101.md) module.
 
 <div class = "care">
 <b style="color: rgb(var(--color-highlight));">A little encouragement...</b><br>
--- a/_module_templates/macros_bash.md
+++ b/_module_templates/macros_bash.md
@@ -1,0 +1,80 @@
+<!--
+
+author:   DART Team
+email:    dart@chop.edu
+version:  1.0.0
+current_version_description: Initial Version
+language: en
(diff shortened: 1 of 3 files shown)


_module_templates/macros_bash.md
```diff
@@ -0,0 +1,81 @@
...
+
+No Previous Versions
+@end
+
+@lesson_prep_bash_basics
+
+You will get the most out of this lesson if you follow along with the examples and try out the commands. In order to do that you need to have a bash shell open on your computer. Please follow the instructions appropriate for the computer you are using.
+
+**Open a bash shell.**
+If you are using a computer with running iOS (i.e. a Mac) you can use the **Terminal** program. If you are on a computer using Windows, open either **WLS** (Windows Linux Subsytem) or **Git Bash**. If you don't have these programs there are instructions for how to download and set them up in the [Bash / Command Line 101](https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/main/bash_command_line_101/bash_command_line_101.md) module.
```
rosemm, line 24:
```suggestion
If you are using a computer with running a Unix operating system (i.e. a Mac or Linux computer) you can use the **Terminal** program. If you are on a computer using Windows, open either **WLS** (Windows Linux Subsytem) or **Git Bash**. If you don't have these programs there are instructions for how to download and set them up in the [Bash / Command Line 101](https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/main/bash_command_line_101/bash_command_line_101.md) module.
```

I believe iOS is only for mobile devices (iphone, ipad, etc.); MacOS is the computer version. It might be clearer to just say "unix" though since that includes Linux builds.

rosemm: I'd also recommend pointing the download instructions to the specific page of the 101 module rather than to its overview

drelliche: You are so right, I copied this directly from the bash 103 module, which obviously needed some updating....

rosemm (changes requested):
Just needs those two little tweaks in the macro, and then let's delete the extra import line from the module itself for tidiness. Otherwise LGTM!

bash_103_combining_commands/bash_103_combining_commands.md
```diff
@@ -20,89 +28,45 @@ After completion of this module, learners will be able to:
...
+@end
 
-The only commands that will be assumed are the navigation commands `cd`, `ls`, and `pwd` and `cat`, all of which are explained in the [Bash / Command Line 101](https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/main/bash_command_line_101/bash_command_line_101.md) module.
+import: https://raw.githubusercontent.com/arcus/education_modules/main/_module_templates/macros.md
 
-**Learning Objectives**
+import: https://raw.githubusercontent.com/arcus/education_modules/main/_module_templates/macros_bash.md
 
-@learning_objectives
+import: https://raw.githubusercontent.com/arcus/education_modules/537-docs-v1-update-bash-combine-commands/_module_templates/macros_bash.md
```
rosemm, line 60:
```suggestion
```
Clever to import both locations like that while drafting! I love it.

rosemm:
Oh wait, we might need to update things in docs as well, because of the new macro -- lemme check.

rosemm:
Will all of our interactive bash modules eventually use that `lesson_prep_bash` macro? Or do you intend it just for some of them?

drelliche:
I expect that there will eventually be higher level bash modules that will not use that lesson prep because they will focus on other things and not need people to download the learning_bash repo

rosemm (approved):

PR #559, opened 2023-08-22
Module development plans and topic wishlist

rosemm:
This is meant to address the following user story: 

Story 1: "As a potential author, I want to understand current needs for new modules, so I can see where my interests overlap with the needs of the project".

Example: Colleagues who work in Johns Hopkins have a lot of materials they could contribute, but they're not sure what our priorities are and where they can make the most impact.  They'd like to do 1-2 modules to start with and wonder where to spend their effort.

Importance: Important.

Possible Solution: Include a link in potential author instructions to a wish list / priorities list that we keep updated.

Feasibility: We need to create a wish list / priorities list for module development so potential authors can know what's underway already and not waste time writing something that's already in development as well as see what modules we're particularly targeting for development.  This documentation can exist within our education_modules repo in some way (in the Wiki? as a file?).  Developing this and maintaining this will require time and a medium complexity of effort.


rosemm:
@paytonk I requested your review on this since you worked on the original user story. Please feel very free to think outside the box with this one -- I thought about a bunch of ways to represent this information (table? GH issue(s)?), and I landed on a plain md doc mostly because that was the easiest thing to implement to get feedback quickly. So definitely push back if you see room for improvement here!

drelliche:
Once this is merged we should also point to it from the public facing website.

pm0kjp (approved):
This is great, TYSM!

PR #564, opened 2023-08-24
Metadata workflow

drelliche:
This is a workflow that will create two new assets under the assets/metadata folder:
1.  a python file which builds the metadata as a pandas dataframe
2. a csv of the same metadata.

This workflow is then implemented as a GitHub action so that it will run anytime someone pushes to main (i.e. merges because main is protected).

This is a janky motorcycle version, so there are of course a few things to be aware of and this might be a reason not to merge this into main yet:
- Currently this is keyed to the metadata_workflow branch, which would have to be changed to main if this is merged into main (3 places in pull_metadata.yml)
- Can the GitHub action even merge into a protected branch? I'm unsure of the answer and may need to investigate further.

These are not reasons to not merge, but ways in which this motorcycle is clearly not quite in racing form:
- Instead of creating an artifact, it runs the bash script twice (about 20 seconds each time) which is cludgy but works
- There is still a problem with `"`s and line breaks in the longer pieces of metadata, currently they are replaced by `+` and `&` to be dealt with on the other end.

To see what this looks like in action, see https://github.com/arcus/education_modules/actions/runs/5968585765

All of that said, merging this will not impact anything public facing, so even if we try to merge and it fails horribly, nothing is particularly broken.


drelliche:
Further investigation about pushing to a protected branch from an action shows that it is HARD (see a [very long discussion](https://github.com/orgs/community/discussions/25305) of this).

I'm going to see if I can set it up to pull the data from `main` but then push the metadata to the `metadata_workflow` branch.

I won't close this PR quite yet because this is a good place for any discussion about the workflow.

drelliche:
New plan: two of these files should be merged into main. The files containing the actual metadata can be merged into main, but they will not automatically update. Instead, once this is in main and triggered by a push to main, it will update the metadata in the metadata_workflow branch, which is not protected. That will be the canonical home of this data.

Before merging:
- in `.github/workflows/pull_metadata.yml line 10 should be changed to "main"

After merging:
- this should be tested to confirm that the files on the metadata_workflow branch do actually update appropriately. 
- Since new metadata has been added since the branches diverged, the workflow should trigger automatically on being merged into main.

pm0kjp (approved):
LGTM

PR #613, opened 2023-08-29
Update macros_r.md

pm0kjp:
This updates the instructions for the R modules, walking people through the process of using the .zip file of our materials, in case they are git-resistant.


drelliche (approved):
seems ok to me!

PR #638, opened 2023-09-04
Updates to reproducibility

### LINKED ISSUE #480: Add wider range of references
record number: 433

Description of issue: user suggest linking to a broader range of outside resources, and specifically mentions psychology as a field with a lot of activity in the area of reproducibility.

### LINKED ISSUE #483: Excel PowerQuery and PowerBI
record number: 447

Description of issue: Learner suggests that we mention that queries in Excel are scripted even though cell formulas are not.

rosemm:
Clarify that excel does support scripts, but it's still worthwhile to learn R or Python
Fixes #483 

Also adds link to a syllabus of academic readings in social science, fixes #480 (Note that it includes repeated use of the word "fuck" though! Note sure if we want to avoid this link for that reason. But it is truly a really excellent list of academic readings, and actually ended up being the basis for a real class I took in graduate school from Prof Srivastava.)


pm0kjp (approved):
Love this, good threading of the needle here

PR #674, opened 2023-09-15
Remove links to lab down skill up webinars by jacob levernier

### LINKED ISSUE #673: Links to Arcus resources don't work
record number: 1323

Description of issue: The links to Arcus resources in the Additional Resources section don't work.

rosemm:
The "lab down skill up" webinar series of videos were all transitioned off of digitalrepository this June. There are new links available via tdnet: https://chop.tdnetdiscover.com/repositories/results?db=299&index=*&fct=Subjects%7Ecommand+line+and+computing+tools%7CSubjects%7Ecommand-line

But those videos are no longer publicly accessible, so we can't use that link in a module. I emailed LS about whether or not we can/should make those links public (the whole video series is possibly being sunsetted this year anyway), but in the meantime this PR removes the reference to those videos since they're inaccessible. Fixes #673 

Note: I searched our whole repo for links to resources on digitalrepository and these two modules were the only ones that came up so I don't think there are any other related links that have the same problem.

### DIFF
--- a/bash_103_combining_commands/bash_103_combining_commands.md
+++ b/bash_103_combining_commands/bash_103_combining_commands.md
@@ -2,10 +2,10 @@
 
 author:   Elizabeth Drellich and Nicole Feldman
 email:    drelliche@chop.edu and feldmanna@chop.edu
-version: 1.3.2
+version: 1.3.3
 current_version_description: Restructured Learning Objectives
 module_type: standard
-docs_version: 1.3.1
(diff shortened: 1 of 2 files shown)


drelliche (approved):
Those were good videos, I hope we regain public access to them!

But for now this is the right call, lgtm!

rosemm:
@drelliche I'll put this content back in, since we have youtube links now! 🎉

PR #677, opened 2023-09-18
Issue template improvement

rosemm:
Adding a `bug` issue template, primarily as a way for us to record some of the common problems and quick fixes we've learned about that are currently only stored in our collective memory. 
(As a reminder, the CONTRIBUTING doc will be automatically linked in a popup by GH whenever someone tries to create their first issue in the repo.)


drelliche (commented):
I have one suggestion about automatically including the record number field even though it might sometimes be irrelevant. 

Other than that, it looks good to me!

drelliche (approved):
lgtm!

PR #686, opened 2023-09-20
Update to docs v2

rosemm:
Brings r reshape long wide up to date with docs v2 (it pretty much already was, but I added the missing metadata field `mode: Textbook` and added one more module to `sets_you_up_for`)


drelliche (approved):
lgtm!

PR #703, opened 2023-09-21
Docs v3

### LINKED ISSUE #718: Add instructions for what grammatical tense to use for `current_version_description` and `version_history` (I vote past tense -- I think it reads better in the overview section that way)

### LINKED ISSUE #719: Add instructions so users know what counts as a sequence or not

rosemm:
This updates docs with the new front matter field, `collection`, and adds text as an option for `data_domain`. Also fixes some typos.
It also includes an issue template for bringing modules up to date with docs v3, but I don't think we'll want or need to  use this template now, since this is only a change in optional front matter that will not impact the way modules render for learners at all. Instead, we'll probably leave this issue template until we are doing more systematic updates (maybe for v4?), and then we can just add this on as an additional check at that point.

A couple little things: Fixes #718 Fixes #719

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -295,19 +295,19 @@ try {
 @add_item(11,import macros_genomics.md)
 @add_item(12,import macros_bash.md)
 
-### `author`
+### author
 
 ```
 author:   Your Name
 ```
 
-### `email`
+### email
 
 ```
 email:    email@example.edu
 ```
 
-### `version`
+### version
 
 ```
 version:  0.0.0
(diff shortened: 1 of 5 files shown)


docs.md
```diff
@@ -652,7 +661,30 @@ Must be one of the following:
...
+- `infrastructure_and_technology`: Focused on software or tools, especially setup and systems. Things like how to install software, or understanding what software and/or languages to use for what tasks. Also includes basic computer skills like navigating file directories.  
+- `intro_to_data_science`: Skills for learners new to data science, including how to troubleshoot, reproducible methods, etc.
+- `learn_to_code`: Modules that is primarily focused on teaching coding. Note that a module might require coding without belonging to this collection, especially more advanced modules that use a coding language to achieve some task (modeling, genomics, etc.) but without the pedagogical focus being on the coding language itself. 
+- `machine_learning`: Modules about machine learning and AI.
+- `statistics`: Modules about statistics, including both applied data analysis (e.g. here's how to do this test in R) and a more theoretical understanding of statistics and the underlying math (e.g. learn the matrix algebra underlying the general linear module).
```
drelliche, line 683:
```suggestion
- `statistics`: Modules about statistics, including both applied data analysis (e.g. here's how to do this test in R) and a more theoretical understanding of statistics and the underlying math (e.g. learn the matrix algebra underlying the general linear model).
```

drelliche (approved):
lgtm!

PR #704, opened 2023-09-21
Add "collection" and other metadata

drelliche:
We collectively decided to add a "collection" category to the module metadata. This PR makes the discussed additions and changes to the metadata of each module and also updated the metadata processing workflow to also collect that new metadata.


rosemm (approved):
Looks great. Thanks so much for doing this!

PR #742, opened 2023-10-12
Add updated links to CLI videos from lab down skill up

rosemm:
Adding links to Jacob Levernier's bash webinars from the lab down skill up series, since we have the links for them on youtube now. Note that Jacob said he'd rather actually not be credited explicitly, so I didn't include his name.


drelliche (approved):
LGTM

PR #743, opened 2023-10-12
Pyodide testing

rosemm:
A few things to note about pyodide in general:

- There are two ways for us to run pyodide code: You can create an executable and editable code block in the page with `@Pyodide.eval`, or you can silently execute python code via pyodide without showing it on the page with `@Pyodide.exec`
- Stuff from previously executed python code stays available when you go to a new page. So loading libraries, saving objects etc. persists throughout the module. 

Pandas module: 

- I added some `@Pyodide.exec` code to load pandas at the top of each page even though it won't be necessary for learners stepping through the module from beginning to end since they'll have loaded it earlier. If someone links directly to a page in the middle of the module, though, this ensures that pandas commands are available even if they drop into the middle without doing the first part of the module first.
- I'm not currently re-loading the `covid_data` DataFrame on every page, since the module involves editing that DataFrame at one point and I'd like the changes to persist. I did re-load it at the beginning of each quiz, though, to ensure learners are working with the correct data as they try quiz questions. Totally open to push back on whether or not we should be reloading it at all, and also whether or not we need some clearer signposting about when/if it's being reloaded.
- We should re-examine the learning objectives and the quizzes to see if we want to change or add formative assessment

### DIFF
--- a/_module_templates/macros_python.md
+++ b/_module_templates/macros_python.md
@@ -68,21 +68,6 @@ If you reload the webpage, however, all of the cells will reset to their origina
 
 @end
 
-@re_import_pandas
-<div class = "help">
-<b style="color: rgb(var(--color-highlight));">Troubleshooting help</b><br>
-
-If you have refreshed this page in your browser or are starting on this page, remember to import the `pandas` package by running:
-
-
-```python
-import pandas as pd
-```
-@Pyodide.eval
-
-
-</div>
-@end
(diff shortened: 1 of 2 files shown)


rosemm:
Testing is done, set up a new branch and PR for actually updating modules with pyodide

drelliche:
This has be subsumed by #849

Issue #768, opened 2023-10-27
avoid printing multiple large tables on a single page


rosemm:
record number: 1769

Description of issue: learner found themselves wanting to scroll up and down to look again at code previously run, but once you run the code it often prints very large tables which makes it harder to scroll on the page without losing your place.

PR #799, opened 2023-11-17
Joy bug squashing

### LINKED ISSUE #767: improve the note about the data being fake
record number: 1769

Description of issue: In the lesson prep section, the example code prints DOB and sex, and then there's an "important" box reassuring learners that the data is fake and saying you can tell because the names and SSN look wrong. But at this point, learners have only seen DOB and sex, both of which look potentially real. Either re-word the important box, or add code to show names and SSN.

pm0kjp:
This fixes #767 and #768

### DIFF
--- a/_module_templates/macros_sql.md
+++ b/_module_templates/macros_sql.md
@@ -2,8 +2,8 @@
 
 author:   DART Team
 email:    dart@chop.edu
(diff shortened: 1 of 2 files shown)


rosemm (approved):
LGTM!

PR #802, opened 2023-11-17
Added highlight box for hint

### LINKED ISSUE #684: Needed to do a loop with a dictionary, but that was never taught
record number: 1407

Description of issue: Learner couldn't figure out how to do this without looking at the answer file because the loop module only showed how to do loops with lists, not dictionaries, so they didn't know how to approach this problem.

leemc-data-ed:
Highlighted a hint for looping through dictionaries to make the content stand out. Merging will close #684.


rosemm (approved):
LGMT!

PR #806, opened 2023-11-20
edit learn_to_code description

### LINKED ISSUE #805: clarify `learn_to_code` collection
edit the description of the `learn_to_code` collection in docs based upon [conversation with Elizabeth](https://github.com/arcus/education_modules/pull/786#discussion_r1385600150) in #786 

New description will better capture @drelliche's intention for the collection, and make it more clear to module authors what does or does not belong in that collection.

franzenr:
resolves #805

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -2,8 +2,8 @@
 
 author:   DART Team
 email:    dart@chop.edu
-version:  3.0.0
-current_version_description: Added collection as a front matter field, added text as a data_domain, fixed typos
+version:  3.1.0
+current_version_description: update definition of learn_to_code collection
 language: en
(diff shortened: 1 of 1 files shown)


franzenr:
@drelliche presumably you'll approve this, since the new description is just copy pasted verbatim from the suggestion you gave on our other thread :) 

I wasn't 100% sure on whether this is minor or a revision, but I went minor since it could technically cause some modules to become out of line with the standards (though I don't think it will, as I think how to troubleshoot is the only one that currently fell into a learning to code gray area). I could potentially be convinced otherwise, though!

drelliche (approved):
LGTM!

PR #830, opened 2023-11-30
Escape underscores

rosemm:
We were having problems with some text with underscores displaying wrong (i.e. italicizing), especially within markdown table cells. I went through and escaped all of the unescaped underscores I could find, even when it wasn't strictly necessary, such as when it was the only underscore in that line. Having them all escaped is more consistent, and hopefully makes our formatting more robust to little changes in how the parser works going forward.


drelliche (approved):
This looks good to me, thank you for adding this consistency!

PR #847, opened 2023-12-07
Workflow improvement

drelliche:
Write metadata to a json for easier portability.


rosemm (approved):
Just a quick review, but I don't see anything that looks likely to break stuff, so I say go for it! :)

PR #848, opened 2023-12-07
Correct front matter checklist so collection is required if both data_task and data_domain are missing

rosemm:
This PR updates docs.md so that the "your frontmatter checklist" section correctly shows `collection` as required if both `data_task` and `data_domain` are unchecked; modules are required to have at least one of `collection`, `data_task`, or `data_domain`.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -244,7 +244,15 @@ You'll need the following fields in your front matter (new fields added by check
 
 @add_item(4,data\_domain)
 @add_item(5,data\_task)
-@add_item(6,collection)
+<script modify="false">
+try {
+  let module_characteristics = @input(`module_characteristics`)
+
+  if(module_characteristics[6] || (!module_characteristics[4] & !module_characteristics[5])) {
+    send.liascript(`- collection 💫`)
+  } else send.clear()
+} catch(e) { }
+</script>
 
 * coding\_required
 
--- a/docs.md
+++ b/docs.md
@@ -25,7 +25,7 @@ try {
 @version_history
 Previous versions: 
 
--[3.0.0](https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/cd0da90a30910c22d5502f509125892a761c3145/docs.md#1): Added collection as a front matter field, added text as a data_domain, fixed typos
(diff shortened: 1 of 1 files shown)


franzenr:
I'm curious if we should be treating `collection` as the default required field, that only is superseded if domain or task is selected? (as we are in the logic implemented here) or if we should be treating all three of them as equally valid "defaults". 

(These might be a totally silly questions to be asking -- I'll be honest that I still don't totally grok the way we're using some of these metadata fields).

Another alternative might be to add an asterisk after each of data domain, data task, and collection, and note that at least one of these must be true for every module?

franzenr:
The only small issue I see with this PR is that the 💫 emoji shows up next to collection even when its appearing in the list by default (that is, "belongs to a collection" is unchecked.) I suppose this could be mildly confusing to someone, or perhaps make them think the logic is broken for some reason, and therefore skip using the collection field, but maybe the most likely impact is just purely aesthetic. 

It suppose it would be much more complicated to only have 💫 appear if "belongs to a collection" is selected, but not if it is only listed because data_domain and data_task are unselected, yeah?

rosemm:
So my thinking with having `collection` as the default if none are checked is that we can easily add collections to cover any topic we need to. If a module doesn't use data directly, it wouldn't be possible to sensibly add a `data_task` or `data_domain` for it. So in the situation where someone is writing a module that doesn't belong to any existing `data_task`, `data_domain`, or `collection` options, they would generally be in the position of adding a new collection. 
Of course, reviewer can push back in QA if they feel something should really have been added as a new `data_task` or `data_domain` instead. But this felt like the most useful way to configure the checklist for authors.

rosemm:
Okay, I've updated it as per your suggestions --- lemme know what you think!

franzenr:
Ah okay, that makes sense to me! And to be clear, I'm fine with that approach over "my" approach, I just was asking since I'm still a bit fuzzy on some of these metadata distinctions. 

So if you think your earlier approach makes more sense for authors, we should totally go that way! 

@rosemm

franzenr (approved):
approving since I'm fine with either, but feel free to revert back to your initial approach @rosemm

PR #855, opened 2023-12-15
pipe quiz flexibility

### LINKED ISSUE #749: Quiz marking correct answer wrong
record number: 1684

Description of problem: User reports entering `mydata %>% select(first_name, last_name)` and it being marked wrong despite being the correct answer.

---

**Some things to try first!** 

- If pyodide cells aren't working as expected in Python modules, we may need to update pyodide. 
- If LiaScript macros aren't rendering properly, especially `overview`, check for blank macros defined in the front matter (e.g.`version_history` is a common culprit). You may need to add load bearing white space. 
- If the `feedback` macro isn't working, check that no front matter fields used in the generated REDCap link are missing (e.g. `module_type`).
- Problems with `div`? You may need a newline after the `div` tag and before the `/div` tag. 
- Quiz answer boxes not displaying properly? There cannot be a newline between the last answer option and the beginning of the question followup section. Make sure the `****` surrounding the answer div starts immediately after the last line of the question.

franzenr:
quiz question about pipes can accept either pipe, whitespace optional
closes #749

### DIFF
--- a/r_basics_transform_data/r_basics_transform_data.md
+++ b/r_basics_transform_data/r_basics_transform_data.md
@@ -1,7 +1,7 @@
 <!--
 author:   Joy Payton
 email:    paytonk@chop.edu
-version: 1.3.4
+version: 1.3.5
 current_version_description: Updated with new metadata and to remove references to Binderhub
 module_type: standard
 docs_version: 2.0.0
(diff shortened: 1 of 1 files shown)


rosemm (changes requested):
This is AWESOME! Just need to increment version.

rosemm (approved):
LGTM!

PR #875, opened 2023-12-21
Add module_id as a front matter field

rosemm:
This PR does a few things: 

- it updates the macros in macros.md to pipe module_id in when creating the redcap survey links (module_id has also been added as a hidden field in that redcap project)
- it adds module_id to each of the templates
- it adds information about module_id to docs.md and increments docs to v4
- it adds a GH issue template for updating modules to docs v4
- it adds a `module_id` front matter field to every existing module and increments version (revision)

FAQ (in my mind): Do we have to update all the module files right now? Couldn't we wait and do that piecemeal? 

No, because as soon as we update macros.md, all of the modules will start using the new definition of the redcap links, which means they'll need a module\_id field to successfully create that link. If we updated macros.md without adding module\_id fields to the modules, we wouldn't be able to collect quick module feedback.


franzenr:
Do we really need to create an issue template for bringing up to speed with docs v4 if we're adding the module_id to all of the fields in this one PR?

rosemm:
Yeah, I wondered that too... I put it in because I was afraid I was overlooking something, some way a module could dodge this update and then get missed? But that probably can't happen. Feel free to delete that file, if you think we should leave it out.

franzenr:
Whipped together fun little script this evening that ensures that every file that is a module has this field, and that the id matches the file name exactly. 

So that should give us assurance that all of the modules actually did get this update! 

There's only one mistake I've caught, which I'm going to fix right now! 

At a future (hopefully soon!) date, we can take the final few steps to make this script run automatically on PRs to main via Actions

rosemm:
Nice! That script sounds awesome! Also, I’m surprised and delighted I got through all those files with only one typo 😊 Can we approve and merge then?

franzenr (approved):

franzenr:
went ahead and also deleted the v4 checklist, which I forgot to to last night! All good to merge now :)

PR #888, opened 2024-02-02
Add version checking workflow

franzenr:
workflow to ensure that module updates  always have their version numbers updated and new modules always have version number of 1.0.0


drelliche (approved):
LGTM! Thanks for the tour of how it works!

PR #893, opened 2024-02-13
expand scope of version check

franzenr:
expand scope of version checking script to look at the other files with the "version:" metadata field. namely, all of the macro files, and docs.md


franzenr:
@drelliche assigned this review to you since you are the person most familiar with this workflow at this point, however if you don't have bandwidth that's totally fine, let me know and I'll ping someone else. 

Should be more or less "trust fall" level, though.

drelliche (commented):
The code looks good to me! 

**Question for the future:** are docs and macros the only places we would ever have version numbers? I don't have even a theoretical example right now, but I'm asking because if we do end up having a version number somewhere else, the name of that script will be confusing because it isn't actually checking for a version number, just for the file being either docs.md or a macro.

**Suggestion**: Maybe it would help to rename `not_module_has_version.sh` to something like `is_docs_or_macro.sh` to help with debugging should you get hit by the lottery :)

franzenr:
love that, that's a way better name anyways. why is naming stuff the hardest part of coding? lol. I'm gonna update now before we merge!!

franzenr:
@drelliche just pushed the changes.

drelliche (approved):
LGTM!

PR #896, opened 2024-02-16
Generalized linear regression

rosemm:
New module!


generalized_linear_regression/generalized_linear_regression.md
```diff
@@ -0,0 +1,408 @@
...
+You're not alone!
+
+For an approachable review of linear regression models, including a review of the equation behind them, check out this [article explaining linear regression](https://education.arcus.chop.edu/ordinary_linear_regression/).
+
+</div>
+
+For example, we could express a model [predicting fetal weight from sonographic measurements of femur length](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6077071/) as the equation:
+
+$$
+fetal\_weight = \beta_0 + \beta_{femoral_length} * femoral_length + e
```
drelliche, line 85:
```suggestion
fetal\_weight = \beta_0 + \beta_{femoral\_length} * femoral\_length + e
```

Another general suggestion (coming from my pedantry in the world of mathematics) is that variables like `femoral_weight` should not be spelled out in "math mode" at all, but that might make them look even worse to people who aren't me. But also that has a lot to do with the LaTex interpreter being used and I don't think I can adequately demonstrate here. 

$$
fetal \_weight = \beta_0 + \beta_{femoral \_length} * femoral \_length + e
$$

might be easier to read as

$$
\text{fetal} \textunderscore \text{weight} = \beta_0 + \beta_{\text{femoral} \textunderscore\text{length}} * \text{femoral} \textunderscore\text{length} + e
$$

Github is rendering this a bit whacky but it sorta does that for all long math lines. I'm not going to dig this hole in my brain any deeper unless you want to talk about it :)

generalized_linear_regression/generalized_linear_regression.md
```diff
@@ -0,0 +1,408 @@
...
+Something like a probability or proportion is bounded at 0 and 1; it's not possible to have a negative probability, or one over 1. 
+
+A variable like age might be unbounded or bounded depending on what ages you're studying. 
+In a young pediatric population, age is bounded at 0 because you can't have any negative ages and a distribution of infant ages will get cutoff at that lower bound. 
+But if you're studying adults, then age might be functionally unbounded in your dataset.
+
+<div class = "learn-more">
+<b style="color: rgb(var(--color-highlight));">Learning connection</b><br>
+
+For an excellent review of continuous vs. categorical variables, see the [Khan Academy video on discrete and continuous random variables](https://www.khanacademy.org/math/statistics-probability/random-variables-stats-library/random-variables-discrete/v/discrete-and-continuous-random-variables).
```
drelliche, line 141: This link seems to be broken.

generalized_linear_regression/generalized_linear_regression.md
```diff
@@ -0,0 +1,408 @@
...
+
+</div>
+
+We can coerce problematic outcome variables into nice, continuous ones so that we can still model them with a line.
+The strategy for transforming your outcome depends on what its distribution is like to begin with --- if it's binary (or bounded at 0 and 1), then a logit works well. 
+If it's count data, then a log is the right transformation, etc.
```
drelliche, line 147: I'd suggest moving some of the technical content of this paragraph later and focus on the first sentence. Explain what it means to "coerce problematic outcome variables" without introducing log and logit yet.

In fact I think I'd move the Logistic Regression intro section up here, maybe with structure like:

Why do we need generalized linear models? (linear only works for unbounded, continuous)
- We can transform/make/coerce bounded or continuous data into a new shape so that linear stuff works again!
- log and logit are the way to do this
- more details for the mathematically interested, which is your current big behind the scenes box.

rosemm: Okie dokie, @drelliche  I reordered things, as per our discussion -- take a look and let me know how you think it's landing now! 
I also tried to tidy up the mathjax, but please be picky and let me know if it's still not great!

generalized_linear_regression/generalized_linear_regression.md
drelliche: @rosemm Apparently liascript doesn't like `\textunderscore` 😢

generalized_linear_regression/generalized_linear_regression.md
rosemm: Do you mind checking this again? This link is working for me...

drelliche: Hmm, it is working for me too now, don't know what was going on before!

rosemm: Cool! I love problems that fix themselves :)

rosemm:
Okay, @drelliche it's ready for you again! I implemented several (hopefully!) improvements: 

- Moved around some existing content to improve the pedagogical flow, as we discussed, especially breaking up the sepsis example so the problematic attempt with the plain linear model occurs early on, and the better attempt with the logistic regression is toward the end
- Smoothed out the new transitions, added a little more scaffolding around, in particular, the plot of the bad linear model
- Substantially reworked the "probability and odds" bit, which is now under the heading "talking about chance" -- lemme know what you think
- Added a whole new quiz to try to more effectively hammer home the question of what is (and isn't) a continuous unbounded outcome, to better support learning objective "identify when a logistic regression model is appropriate"
- Added a little more explanation of plain linear models at the very beginning. I'm happy to take that out again, though, if you think it's overkill given that OLS is listed as a prereq.

drelliche:
@rosemm This looks great! I'll go over it for all of the checks/typos, etc in the next few days.

generalized_linear_regression/generalized_linear_regression.md
```diff
@@ -0,0 +1,528 @@
...
+Gaussian regression is actually just regular linear regression! 
+A Gaussian distribution, also called a [normal distribution](https://www.khanacademy.org/math/statistics-probability/modeling-distributions-of-data/normal-distributions-library/a/normal-distributions-review), is the [default assumption for linear regression models](https://www.theanalysisfactor.com/assumptions-of-linear-models/). 
+So it's actually no transformation at all!
+
+</div>
+***
+
+## Additional Resources
+
+There's an excellent chapter on [Logistic Regression](https://online.stat.psu.edu/stat504/lesson/6) published by Penn State.
```
franzenr, line 522:
```suggestion
There's an excellent [chapter on Logistic Regression](https://online.stat.psu.edu/stat504/lesson/6) published by Penn State.
```

generalized_linear_regression/generalized_linear_regression.md
```diff
@@ -0,0 +1,528 @@
...
+
+If you'd like to compare output from different kinds of statistical software, [UCLA's IDRE](https://stats.oarc.ucla.edu/other/dae/) has published data analysis examples for logistic regression in [R](https://stats.oarc.ucla.edu/r/dae/logit-regression/), [Stata](https://stats.oarc.ucla.edu/stata/dae/logistic-regression/), [SPSS](https://stats.oarc.ucla.edu/spss/dae/logit-regression/), [SAS](https://stats.oarc.ucla.edu/sas/dae/logit-regression/), and [MPlus](https://stats.oarc.ucla.edu/mplus/dae/logit-regression/).
```
franzenr, line 524: I know its a bit clunkier, but this would provide much better context for folks using screenreaders

```suggestion
If you'd like to compare output from different kinds of statistical software, [UCLA's IDRE](https://stats.oarc.ucla.edu/other/dae/) has published data analysis examples for logistic regression in a variety of softwares: 
* [Logistic Regression in R](https://stats.oarc.ucla.edu/r/dae/logit-regression/) 
* [Logistic Regression in Stata](https://stats.oarc.ucla.edu/stata/dae/logistic-regression/) 
* [Logistic Regression in SPSS](https://stats.oarc.ucla.edu/spss/dae/logit-regression/)
* [Logistic Regression in SAS](https://stats.oarc.ucla.edu/sas/dae/logit-regression/)
* [Logistic Regression in MPlus](https://stats.oarc.ucla.edu/mplus/dae/logit-regression/). 
```

generalized_linear_regression/generalized_linear_regression.md
```diff
@@ -0,0 +1,528 @@
...
+<div class = "behind-the-scenes">
+<b style="color: rgb(var(--color-highlight));">Behind the scenes</b><br>
+
+It's not necessary to really understand the logit transformation to be able to use logistic regression, so feel free to skip this bit if you don't find it helpful. 
+But if you'd like to get a better idea of how the math works, here goes!
+
+Data bounded at 0 and 1, like probabilities or binary outcomes, can be converted to unbounded data as follows:
+
+1. Probabilities can be converted to odds by dividing the probability by 1-(the probability). So a probability of .5 becomes $0.5/(1-0.5) = 1$.
+2. Odds are unbounded at the upper end (they can theoretically go to positive infinity), but they're still bounded at the lower end, at 0. We can fix that by taking the natural log of the odds. The natural log of the odds is called a logit. For a probability of .5, the logit would be $Ln(0.5/1-0.5) = 0$.
```
franzenr, line 351:
```suggestion
2. Odds are unbounded at the upper end (they can theoretically go to positive infinity), but they're still bounded at the lower end at 0. We can fix that by taking the natural log of the odds. The natural log of the odds is called a logit. For a probability of .5, the logit would be $Ln(0.5/1-0.5) = 0$.
```

generalized_linear_regression/generalized_linear_regression.md
```diff
@@ -0,0 +1,528 @@
...
+
+Logistic regression (regression using a logit link function) works well for binary outcomes.
+But what about other kinds of outcome variables?
+
+If your outcome is counts of something, you can use a **Poisson** or **Negative Binomial** link function. 
+For example, perhaps your outcome is the number of [Adverse Childhood Experiences (ACEs)](https://www.cdc.gov/violenceprevention/aces/fastfact.html) per patient. 
+It's not possible for someone to have fewer than 0 ACEs, so that variable is bound at 0. 
+It's also measured as a count (how many total ACEs experienced), so no one will have 1.2 ACEs, for example. 
+
+For outcomes bounded at 0 and 1 (like a probability or proportion), you can use logistic regression, but a **beta** link function may be more appropriate.
```
franzenr, line 435: I was initially confused by this -- since I didn't immediately connect that this doesn't include our binary variables, since those, too, are bounded at 0 and 1. Is it correct to say that here you're referring specifically to variables that are bounded but continuous within those bounds?

generalized_linear_regression/generalized_linear_regression.md
```diff
@@ -0,0 +1,533 @@
...
+
+</div>
+
+If your outcome is positively skewed, especially something bound at 0, then you can use a **Gamma** or **Inverse-Gaussian** link function. 
+These are especially useful for outcomes like cost, income, or length of stay -- continuous variables bound at 0 and usually positively skewed.
+
+<div class = "learn-more">
+<b style="color: rgb(var(--color-highlight));">Learning connection</b><br>
+
+To learn more about how to talk about the shape of a variable, such as identifying positive and negative skew, see the [Khan Academy lesson on shapes of distributions](https://www.khanacademy.org/kmap/measurement-and-data-j/md231-data-distributions/md231-displays-of-distributions/v/shapes-of-distributions).
```
franzenr, line 450: Not a review suggestion, I just thought you might find this delightful: 

my research stats prof in undergrad taught us to remember how to talk about skew by thinking about the fact that it's named after the outliers that "ruined" the normal distribution. Like ["I would have gotten away with being normally distributed too, if it weren't for you meddling outliers"](https://scoobydoo.fandom.com/wiki/List_of_%22And_I_Would_Have_Gotten_Away_With_It_Too,_If_It_Weren%27t_For_You_Meddling_Kids%22_Quotes) 

Of course, I promptly filed this mnemonic device away in my head under the name "Skew-by Doo".

generalized_linear_regression/generalized_linear_regression.md
```diff
@@ -0,0 +1,533 @@
...
+
+</div>
+
+## Special Considerations for Generalized Linear Models
+
+When you run generalized linear models, it feels very similar in many ways to running a regular linear model. 
+Your output will look largely similar (coefficient tests, measurement of overall model fit, etc.). 
+Under the hood, though, your statistical software is doing something very different.
+
+Unlike linear models, there is not a single analytic solution to generalized linear models -- that means you couldn't (even if you wanted to) figure out the solution by solving the equation by hand.
```
franzenr, line 460:
```suggestion
Unlike linear models, there is not a single analytic solution to generalized linear models -- that means you couldn't figure out the solution by solving the equation by hand, even if you wanted to. 
```

generalized_linear_regression/generalized_linear_regression.md
```diff
@@ -0,0 +1,533 @@
...
+
+When you run generalized linear models, it feels very similar in many ways to running a regular linear model. 
+Your output will look largely similar (coefficient tests, measurement of overall model fit, etc.). 
+Under the hood, though, your statistical software is doing something very different.
+
+Unlike linear models, there is not a single analytic solution to generalized linear models -- that means you couldn't (even if you wanted to) figure out the solution by solving the equation by hand. 
+Instead, the computer uses brute force to find a solution, iterating over tons of combinations of parameter estimates and seeing which gives the best model fit.
+
+This means the results are not exact --- they depend on the sampling algorithm being used --- so you might see very small changes if you run the model again, or if you run it on other software. 
+There's also a risk that your model won't converge if it's too complicated and you don't have enough data.
```
franzenr, line 464: Do we need to insert a brief sidebar about what it means for a model to converge (or, alternatively just use simpler phrasing here)? Its such a brief note that I could be convinced this is fine, but wanted to flag that it could be a bit too jargony for someone who doesn't run many models.

generalized_linear_regression/generalized_linear_regression.md
```diff
@@ -0,0 +1,533 @@
...
+
+True or False: Generalized linear models actually still use a line for the model (like ordinary linear regression), they just transform the outcome variable first.
+
+[(X)] TRUE
+[( )] FALSE
+***
+<div class = "answer">
+
+True! 
+We can only use ordinary linear regression when the outcome variable is continuous and unbounded, so if you have, for example, a binary outcome, then a regression model won't work unless you first transform the outcome to be continuous and unbounded -- that's what generalized linear models do!
```
franzenr, line 476:
```suggestion
We can only use ordinary linear regression when the outcome variable is continuous and unbounded. So if you have, for example, a binary outcome then a regression model won't work unless you first transform the outcome to be continuous and unbounded -- that's what generalized linear models do! 
```

generalized_linear_regression/generalized_linear_regression.md
```diff
@@ -0,0 +1,533 @@
...
+[(X)] A link function is a transformation to convert an outcome variable to a form that is appropriate for linear regression
+***
+<div class = "answer">
+
+A link function is a transformation to convert an outcome variable to a form that is appropriate for linear regression.
+($b_0 + b_x * x$ is an example of a linear model, not a link function.)
+
+The transformation that will work depends on the shape of the outcome variable to start with. 
+For binary outcomes, a logit transformation works well. 
+A generalized linear model with a logit link function is often called "logistic regression" for short.
```
franzenr, line 493:
```suggestion
```
this gives away the answer to the next question.

generalized_linear_regression/generalized_linear_regression.md
```diff
@@ -0,0 +1,533 @@
...
+- the equation of a line (intercept and slope)
+@end
+
+@learning_objectives  
+After completion of this module, learners will be able to:
+
+- describe how generalized linear models are similar to and different from linear models
+- define a "link function"
+- identify when a logistic regression model is appropriate
+@end
```
franzenr, line 35: This should probably also include a LO about identifying boundedness and continuity-ness  of variables! you've already got the quiz questions to cover that content, so I don't think it would be any change other than adding here.

franzenr (changes requested):
Looking great! Just a few small suggestions, and a point or two of clarification :)

generalized_linear_regression/generalized_linear_regression.md
```diff
@@ -0,0 +1,528 @@
...
+If you want to model a categorical outcome, or one that's bounded (like a proportion or percent), then you can't rely on regular linear models to get the job done. 
+
+For example, let's return to the linear model example we were considering, predicting heart rate from temperature in febrile children. 
+As a reminder, here's the linear equation for that model: 
+
+$$
+\text{heart} \_ \text{rate} = \beta_0 + \beta_{\text{temp}} * \text{temp} + e
+$$ 
+
+But what if you're interested in predicting not heart rate, but whether or not the patient is at risk for sepsis, a [binary](https://en.wiktionary.org/wiki/binary) (BI-na-ree) outcome?
```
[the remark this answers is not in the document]
rosemm: The pronunciation I put in is closer to the one in the linked wiktionary page, so I'm guessing it might be more standard than bi-nair-ee (although I feel like that's closer to what I say, too). I think we should keep as-is.

generalized_linear_regression/generalized_linear_regression.md
rosemm:
```suggestion
There's also a risk that your model won't converge (meaning you won't get any results at all) if it's too complicated and you don't have enough data. 
```

Not anything close to a definition or explanation of convergence, but I'm trying not to get myself into the weeds here :) Does this seem like a good balance between clarity and brevity?

generalized_linear_regression/generalized_linear_regression.md
```diff
@@ -0,0 +1,533 @@
...
+
+Logistic regression (regression using a logit link function) works well for binary outcomes.
+But what about other kinds of outcome variables?
+
+If your outcome is counts of something, you can use a **Poisson** or **Negative Binomial** link function. 
+For example, perhaps your outcome is the number of [Adverse Childhood Experiences (ACEs)](https://www.cdc.gov/violenceprevention/aces/fastfact.html) per patient. 
+It's not possible for someone to have fewer than 0 ACEs, so that variable is bound at 0. 
+It's also measured as a count (how many total ACEs experienced), so no one will have 1.2 ACEs, for example. 
+
+For outcomes bounded at 0 and 1 (like a probability or proportion), you can use logistic regression, but a **beta** link function may be more appropriate.
```
rosemm, line 435:
```suggestion
For continuous outcomes bounded at 0 and 1 (like a probability or proportion), you can use logistic regression, but a **beta** link function may be more appropriate.
```

generalized_linear_regression/generalized_linear_regression.md
```diff
@@ -0,0 +1,528 @@
...
+[( )] neither continuous nor unbounded
+***
+<div class = "answer">
+
+Length of stay is continuous because it can be any value in its range -- a patient might have a measured length of stay of 2 days, or 2.5 days, or 2 days 3 hours and 12 minutes, etc. 
+It's not discrete categories of time, it's a continuous scale.
+
+But it's not unbounded! 
+Although there's no theoretical upper limit on length of stay, it would be impossible to have a length of stay less than 0 days, so the variable is bound at zero.
+And unless Dr. Jemison is studying a particular population that all have unusually long hospital stays, it's likely that the data will be effectively cutoff at 0 -- there will plenty of patients with a length of stay of 3 days, and 2 days, and 1 day, but then of course nothing below zero.
```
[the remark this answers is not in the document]
rosemm: I've tried to include a few examples in the module to emphasize that boundedness is sometimes a subjective judgment -- e.g. I talk about IQ as well, which is also technically bound at 0, but in reality never gets close to that cutoff so it behaves as though it were unbounded, and we generally treat it so statistically. I also talk about being potentially bounded, if you're looking at a pediatric population where the 0 cutoff will affect the data, but unbounded if you're looking at an older population where the youngest participants are well above 0. 
It's a slippery issue, so I do want to make sure to treat it thoroughly. This mention is the third example, though, so maybe that's enough?

generalized_linear_regression/generalized_linear_regression.md
rosemm: Elizabeth and I beefed up a lot of the boundedness/continuity material with the idea that it was supporting LO 3, identify when logistic regression is appropriate (and, to a lesser extent, LO 1). 
We could add another LO for it, but I think my inclination is not to, as I doubt that those things per se will feel valuable to learners (I don't think anyone looking at our modules will be like how can I learn about "boundedness and continuity?" It's just that they actually really need to understand both of those things to know when they need to move to a generalized linear model). Since the LOs are for the overview, I think they're more useful when they speak to skills and topics a potential learner might already be able to parse, without having done the module first.

generalized_linear_regression/generalized_linear_regression.md
rosemm: I think I addressed this, lemme know if you still find it confusing.

rosemm:
Thanks for the great edits, @franzenr! I think it's ready for you again -- lemme know if you want further clarification or anything :)

generalized_linear_regression/generalized_linear_regression.md
franzenr: Love it!

generalized_linear_regression/generalized_linear_regression.md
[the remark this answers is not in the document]
franzenr: ok, fine by me!

generalized_linear_regression/generalized_linear_regression.md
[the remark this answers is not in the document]
franzenr: What you've said here makes sense to me, and I agree with the importance of giving this a thorough treatment but I'm still confused by the sentence "unless Dr. Jemison is studying a particular population that all have unusually long hospital stays, it's likely that the data will be effectively cutoff at 0" -- why does the fact that the stays will be short in length indicate that the data will be effectively cutoff at zero? 

Maybe I just got stuck in a weird thought loop here that explains my confusion, but it seems to me that the fact that its cutoff at 0 (as a lower bound) is already explained by the fact that you can't have a negative number of days spent in the hospital. This is true regardless of if she's studying a population with relatively short or long duration of stay. 

If she's studying mostly patients with a short duration of stay, then maybe there's some effective upper bound -- which is the conclusion the first half of that sentence + the portion after the `--` brings me to. Am I on track to say that's what you're intending to get at? 
Where I get caught up is in the middle of the sentence where it says that the data is bound at zero effectively, because she's not looking at pts with abnormally long lengths of stay. 

Another way to put this is that even if she was studying patients with very long duration of stay (1 month or more, say), the data would still be cut off at zero at the lower bound, right?

generalized_linear_regression/generalized_linear_regression.md
franzenr: That one extra word makes all the difference! LGTM!

generalized_linear_regression/generalized_linear_regression.md
franzenr: Hmm, I see where you're coming from! 
The hang up I have (and reason I brought this up) is the QA rule stating "All major topics covered are represented by a learning objective -- there's no significant knowledge imparted that isn't specified in the learning objectives" -- continuity and boundedness seem like significant bits of knowledge, especially given the extent to which we cover them. 

Do you think we might, in the future, write other modules in  in which knowing how to determine boundedness and continuity might be a pre-req? If so, I'd lean toward thinking that having it as a LO might be helpful? 

That being said, I know some of the QA things are clear cut (like the presence of alt text), and others (like this one) are a bit more subjective, so I could be convinced that this shouldn't be a 4th LO if even after re-reading that QA line item you're still not convinced it makes sense to add :)

franzenr:
@rosemm I've got two replies for you to mull over, but that's it!

generalized_linear_regression/generalized_linear_regression.md
rosemm: Re whether continuity/boundedness could come up as a future pre-req, that's a great question... I'm think I'm leaning towards "no", at least not in the way they're treated here -- this discussion is really tightly calibrated to how features of your outcome variable impact whether or not you can use a linear model. For other applications (like when to run an ANOVA, or whether to do a Welch approximation for your t-test, etc.), we'd want a more general discussion of different kinds of variables (and for that I generally link to Kahn Academy lessons anyway -- I don't think we need to write that content). I don't think the way it's explained here would really be particularly useful outside of potential future modules on linear models or generalized linear models, and in that case the relevant pre-req would be something much closer to the LOs that are already articulated here, especially LO 1.
Re the QA requirement: If I were to add an LO for continuity and boundedness, I think I'd suggest something like "identify whether or not a variable can be treated as continuous and unbounded", but for me that is basically the first half of "identify when a logistic regression model is appropriate" (in my head the GLM decision chart is something like "is your outcome continuous and unbounded? yes, use OLS. no, then is it binary? yes, then use logistic regression, no then look at other GLMs." :) )
That said, if you still feel like the module would benefit from LO 4, then let's put it in! :)

generalized_linear_regression/generalized_linear_regression.md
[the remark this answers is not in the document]
rosemm: Ah, oh, I see! I think maybe the culprit is my choice of the word "effectively" . Length of stay data are *technically* always cutoff at zero, as you note, because negative stays make no sense. But in a population that all have long stays, the data aren't cutoff anywhere in a way that means anything practically, they just peter out at the high and low ends. So maybe data from a long term care facility, where all the patients are there at least a month or two, some maybe for years, the data are effectively unbounded. Even though technically there is a bound at 0, it's just that we'd never hit it so the fact that it's technically there doesn't matter statistically. 
But most length of stay data are bounded at zero in a way that does matter, because there tend to be plenty of observations right up until that bound at 0. 
Does this word change help?

```suggestion
And unless Dr. Jemison is studying a particular population that all have unusually long hospital stays, it's likely that the data will be noticeably cutoff at 0 -- there will plenty of patients with a length of stay of 3 days, and 2 days, and 1 day, but then of course nothing below zero.
```

rosemm:
Thanks! Back to you :) 🎾

generalized_linear_regression/generalized_linear_regression.md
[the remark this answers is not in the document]
franzenr: AH okay, I see where you were going with this now!!!! 

Maybe situating it in relation to some of those examples where we can treat it as unbounded would help? 

Here's my iteration! Does this still emphasize the correct point? 

```suggestion
And unless Dr. Jemison is studying a particular population that all have unusually long hospital stays, it's likely that the data will actually run right up to the boundary at 0 -- there will plenty of patients with a length of stay of 3 days, and 2 days, and 1 day, and then of course nothing below zero. This is why we would want to treat it as bounded statistically, unlike IQ, which you may recall we can treat as unbounded despite also not allowing for negative values, because in practice no IQ measurements are ever close to the scale boundary of 0. 
```

generalized_linear_regression/generalized_linear_regression.md
franzenr: Hmm, good points! 

I think I'm convinced we don't need to add it!

franzenr:
one last(?) lob @rosemm 🎾

generalized_linear_regression/generalized_linear_regression.md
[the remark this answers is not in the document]
rosemm: I like this approach! What do you think about keeping the explanation closer to the current example, though, as below?

```suggestion
And unless Dr. Jemison is studying a particular population that all have unusually long hospital stays, it's likely that the data will actually run right up to the boundary at 0 -- there will plenty of patients with a length of stay of 3 days, and 2 days, and 1 day, and then of course nothing below zero.
That's why length of stay would have to be treated as bounded in this example.

If Dr. Jemison was in fact studying a population with very long hospital stays (perhaps patients with a condition that nearly always requires hospital stays of weeks or months) and none of the lengths of stay in the data were close to 0, then it could be treated as unbounded -- even though there would still technically be a lower bound at 0, that doesn't matter statistically if the data don't approach it (just like in the IQ example earlier). 
```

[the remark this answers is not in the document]
franzenr: Love it!!!!

franzenr (approved):
Great work Rose! This is an awesome addition to our module collection :)

PR #899, opened 2024-02-27
Alt text checker

franzenr:
Created a new automation that will peer into our markdown files and make sure we never leave completely empty alt texts.

### DIFF
--- a/.github/scripts/alt_text.sh
+++ b/.github/scripts/alt_text.sh
@@ -1,4 +1,4 @@
-grep -rn --include=\*.md "\!\[\]"  | # search for lines with "![]" empty alt tag, only in markdown files
+grep -rn --include=\*.md "\!\[\s*\]"  | # search for lines with "![]" empty alt tag, only in markdown files
 grep  -v "ISSUE_TEMPLATE" > issues #return only lines in files that are NOT in the issue template folder
  #bc QA issue templates include example code block showing empty tag for illustration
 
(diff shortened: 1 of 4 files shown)


.github/scripts/alt_text.sh
```diff
@@ -0,0 +1,10 @@
+grep -rn --include=\*.md "\!\[\]"  | # search for lines with "![]" empty alt tag, only in markdown files
```
drelliche, line 1: Would it be just a few more regex characters to also check for alt tags that have empty spaces? I think this might do it...
```suggestion
grep -rn --include=\*.md "\!\[\s*\]"  | # search for lines with "![]" empty alt tag, only in markdown files
```

drelliche: I bring this up because I suspect that I will be an offender of this error type and I wouldn't want to let myself accidentally evade the checker :)

.github/workflows/alt_text.yml
```diff
@@ -0,0 +1,13 @@
...
+    branches: [ "main" ]
+    types: [opened, synchronize]
+
+jobs:
+  check-alt-text:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Check markdown for alt text presence 
+        run: bash .github/scripts/alt-text.sh
```
drelliche, line 13: I only noticed this because it failed itself!

```suggestion
        run: bash .github/scripts/alt_text.sh
```

drelliche (commented):
One typo and one suggestion, then it looks great!

.github/workflows/alt_text.yml
franzenr: whoops lol good catch!

.github/scripts/alt_text.sh
franzenr: excellent call, I didn't even think about that as a fail case to include! Thanks for calling it out :)

drelliche (approved):
LGTM!!!

PR #905, opened 2024-03-14
Git and bash metadata updates

### LINKED ISSUE #890: Bash and Git depends on new CLI module
The Bash and Git sequences have dependencies on the new Demystifying the CLI module that should appear in its metadata. Figure out what those are and add them.

leemc-data-ed:
Added Demystifying the Command Line Interface to a bunch of module metadata as a pre req and/or in "depends on knowledge available in" list. Would close #890.


drelliche (approved):
LGTM!

PR #906, opened 2024-03-15
update marketplace actions to v4

franzenr:
update download-artifact, upload-artifact, and checkout to use v4, to address deprecation of node.js 16 actions


drelliche (approved):
lgtm, let's give a spin!

PR #925, opened 2024-04-29
Remove version reconstruction

franzenr:
remove instructions for reconstructing version history, as we have completed that initiative.


leemc-data-ed (approved):
Agree that this seems like a revision. Lgtm!

PR #927, opened 2024-04-30
Move learning_objective boiler plate to overview and feedback macros

rosemm:
I moved the text "After completion of this module, learners will be able to:" out of each module's individual `learning_objectives` macro and into the `overview` and `feedback` macros, which is where  the LOs are currently printed. 

This PR is a minor change in the sense that there are no edits that will affect learners' experience of modules, but it is a major change in that it touches every published module file, as well as the `overview` and `feedback` macros, docs.md, and all three templates. 

Note that we informally discussed maybe changing the wording before printing LOs, especially in the feedback section to something other than "After completion of this module, learners will be able to:" but that is not part of this PR. In future, we can make changes like that more easily, since it will be an edit in just macros.md rather than needing to edit each module file.


drelliche (approved):
Looks good to me!

PR #929, opened 2024-05-02
Add #1 to links that need it

pm0kjp:
Updated links that point to liascript content to make sure they point to a specific page.


drelliche (commented):
I don't know if we need the changes to the `current_version_description`s since it is a revision, especially for modules that are still in version 1.0.x, but I'm not strongly opposed to them.

The metadata files are fine, but also irrelevant as they aren't the source of truth for the metadata, that is on a different branch: https://github.com/arcus/education_modules/tree/metadata_workflow

pm0kjp:
> I don't know if we need the changes to the `current_version_description`s since it is a revision, especially for modules that are still in version 1.0.x, but I'm not strongly opposed to them.
> 
> The metadata files are fine, but also irrelevant as they aren't the source of truth for the metadata, that is on a different branch: https://github.com/arcus/education_modules/tree/metadata_workflow

The "version description" is something I went back and forth on, but finally decided that it makes sense to track some level of what changed, especially so 1.0.3 of something isn't later listed with just "initial version" -- because, uh, what did those three revisions accomplish?  It's helpful to us and to our learners, I think, to include details about typo fixes, etc. in case the numbering doesn't otherwise make sense.

drelliche (approved):
lgtm

PR #934, opened 2024-05-22
add link check workflow

franzenr:
periodically check for any links that no longer work in our module content. 

------
A few notes: 

Because there are files that always cause some failures (e.g., the intentionally bad urls being used as an example in a regex module), any errors found in a few files (how_to_troubleshoo.md, |macros.md, macros_python.md, and demystifying_regular_expressions.md) are excluded from the report that gets generated. This means this workflow will not catch any actually bad links in those modules. For a later iteration, we can always refine the code to not exclude those files all together, and instead just remove mentions of the links that we expected to fail, but this feels like a good skateboard. 

Also, while the cron syntax is valid, I'm not 100% sure it will accomplish what I actually want for it to accomplish (which is running only once every two weeks on a Tuesday, so we run it just before either backlog refinement or sprint planning). I don't feel like waiting to find out in a different repo though, and the worst consequences of it not working as expected are either (a) we get annoyed by way too many issues being created or (b) I realize in a month that it wasn't running when I wanted it to. These both seem like they are fine to happen on production. 

LMK if you have any questions or want me to walk you through the code in more depth or think the file needs more verbose commenting

### DIFF
--- a/.github/workflows/link_check_cron.yml
+++ b/.github/workflows/link_check_cron.yml
@@ -39,4 +39,4 @@ jobs:
         with:
           title: Link Checker Report
           content-filepath: issue_content.md
-          labels: report, automated issue, bug
+          labels: bug
--- a/.github/workflows/link_check_cron.yml
+++ b/.github/workflows/link_check_cron.yml
@@ -1,0 +1,42 @@
+name: Link check cron job
+
+# the role of this link check workflow is to periodically look for links
+# that have died since last QA. If broken links are found, action still passes
+# and an issue is created
+
+on:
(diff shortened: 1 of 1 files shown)


rosemm (approved):
This sounds great! I it looks like a lot of the work here is being done by `lycheeverse/lychee-action@v1.9.0` and `peter-evans/create-issue-from-file@v5`, and I didn't go track those down to see how they work, but I think that's fine -- the worst case scenario here is that we'll either have issue created that we don't want/need, or the issues we expect won't be created, and neither of those is a big problem. 
Thanks so much for putting this together! I think it will be so helpful!

PR #943, opened 2024-07-11
update directory structure requirements

franzenr:
data folder technically new to docs, but already in use. 

Also, changed `src` to `code` because `src` is not being used by anyone yet (mostly `notebooks` is in use, but that isn't well named for non-notebook files), so it will require the same amount of edits to bring up to standards either way, and `code` is more self-explanatory for learners who might poke around in our repo on their own to try to find materials. 

Made edits to the QA templates for now, though won't be long until we can deploy the action that looks for this, at which point we should be able to just remove that section from the manual QA process all together.


rosemm (approved):
LGTM! I added a checklist issue template for updating modules to v5

PR #956, opened 2024-12-18
Add fix for parser, changing internal markdown to html

pm0kjp:
Old (broken) cartoon in last box: https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/main/demystifying_large_language_models/demystifying_large_language_models.md#8
New (fixed with help from Andre): https://liascript.github.io/course/?https://raw.githubusercontent.com/arcus/education_modules/fix-cartoon/demystifying_large_language_models/demystifying_large_language_models.md#8

The new parser breaks nested markdown, Andre provided this fix, we might need to copy this approach elsewhere.  He writes:  

> One " was not closed, I added one after DeepAIHire ... I will have to fix this, the parser seems to be too greedy in this case, actually it should stop at the closing bracket ]. The quotes are translated into Unicode symbols, based on the language definition. You will see the difference better, if you change the language to de (German), we start at the bottom, or fr who use << >> ... but this should not affect the link detection ...
> Parsing titles in images is a simple rule, that searches for the closing brackets and parentheses " ... ) only. In this case nested links, will be identified as the end. I changed this to html-links ... but I will also try to make this more robust

### DIFF
--- a/demystifying_large_language_models/demystifying_large_language_models.md
+++ b/demystifying_large_language_models/demystifying_large_language_models.md
@@ -2,7 +2,7 @@
 module_id: demystifying_large_language_models
 author:   Joy Payton
 email:    paytonk@chop.edu
-version:  1.0.6
+version:  1.0.7
 current_version_description: Initial version
 module_type: standard
 docs_version: 2.0.0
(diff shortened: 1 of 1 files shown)


leemc-data-ed (approved):
Lgtm!

PR #958, opened 2025-02-07
Feedback macro

franzenr:
Moved feedback link up to top of page, and also turned it into a button. 

I'm honestly not sure how I feel about it as a button, or if I think it would be better as just a link (which I implemented first, in [this commit](https://github.com/arcus/education_modules/commit/0116eac066f1ec00f3588b0576b3933decd9cbe7), screenshots below).

Version of feedback page using button

Version of feedback page using link just moved to top of page


rosemm (approved):
I love it!

PR #963, opened 2025-03-17
add missing instructions in quiz q, resolves #962

### LINKED ISSUE #962: Missing instructions in Quiz
record number: 2109

Description of issue: The first question on the quiz doesn't clarify that you should select all that apply, though the other two questions do (the question does indeed have more than one correct answer).

franzenr:


pm0kjp (approved):
LGTM!  Thanks for quick fix!

PR #969, opened 2025-09-08
Database normalization quiz tweak

rosemm:
Reframing the one-to-many quiz question to ask about one-to-one rather than one-to-many since the "correct" answer is actually that patient to provider and provider to patient are both "many-to-many" rather than "one-to-many". By framing the question in terms of identifying which relationships are one-to-one, we avoid that ambiguity.


leemc-data-ed (approved):
Lgtm!

PR #974, opened 2025-11-14
hide quiz explanations when questions not answered yet

rosenfrantz:
Hi friends! @leemc-data-ed @rosemm Thanks to the power of open source, I was able to share some of our materials with learners at my new job who needed some additional resources on Python. And, thanks to the power of open source, I can directly fix an issue I noticed while looking over them again! Yay open source!!!!! 🎉

something must've changed in LiaScript, and now quiz explanations seem to not be playing nicely with indentations.

Removed indentations so that question explanations are only shown after correct answer or learner clicking to show solution. The indentation did make it look marginally nicer with the numbered questions, but this seems like the simplest solution, and its not ugly by any means. 

I did a quick grep on the repo to look for the pattern, and these 3 are the only affected modules I could locate!

### DIFF
--- a/python_basics_lists_dictionaries/python_basics_lists_dictionaries.md
+++ b/python_basics_lists_dictionaries/python_basics_lists_dictionaries.md
@@ -2,7 +2,7 @@
 module_id: python_basics_lists_dictionaries
 author:   Meredith Lee
 email:    leemc@chop.edu
-version: 1.2.2
+version: 1.2.3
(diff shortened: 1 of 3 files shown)


rosenfrantz:
oops, I forgot about updating the versions -- thankfully someone really cool and smart and beautiful had the idea to write an action that enforces a check for that on all pull requests!

rosemm (approved):
Yay!

PR #975, opened 2026-01-26
Bug fixes

leemc-data-ed:
Plots are accumulating as the module progresses; each code cell prints all previous plots as well as the intended one. Adding `plt.close` to the plot-generating code cells fixes this issue. Also added a line of explanation in the text.


drelliche (approved):
What a frustrating bug! Good fix!

Increment the version number and this is great!

